PR #4, opened 2017-05-02
Feature/no redis option

aseigo:
In order to consider an alternative to Redis for storage / notification, gathering the serialization code together into one place: the Serializer module (now called Serializer.Redis). Previously, these functions were sprinkled across the various modules, but that obviously does not "scale" should other (non-compatible, in terms of serialization) mechanisms are added. It is perhaps also cleaner to put all the serialization code together anyways?


tompave:
Thank you for the PR.
These days I'm a bit busy, but I think these changes make sense.

I'm warming up to the idea of including first-class support for multiple adapters, and I need to think about how the internal API should change to make it easier to add them.

aseigo:
> These days I'm a bit busy,
No problem; I think this is probably true for most of us, after all! :)

> but I think these changes make sense.
Cool..

>  need to think about how the internal API should change to make it easier to add them
I'll see what I run into (if anything?) while implementing a simple ecto-for-storage and global-process-for-notification combo and provide feedback as that progresses .. 

Cheers!

tompave:
Hi @aseigo, thank you, but I must point out that with these new commits the changes have become quite extensive. I don't think I can just accept this sort of PR, because first I really want to think about how to support any storage and notification adapters (should they always come together? Can they be independent? etc)

I would be more inclined to merge the original content of the PR: the simple extraction of the Redis related serialization functions, so that everything is inside `Serializer.Redis`. As you very reasonably pointed out, it would make the library tidier.

What you are doing here looks like great work, however, and I can see how it makes sense. If you want to continue working in this direction in your fork that's great, but I want to be realistic about the chances of the PR to be merged as it is right now.

aseigo:
Yes, that's fine; I forgot that github automatically updates PRs when pushing to a branch. These changes should be brought in incrementally (or .. if it all .. that's up to you, obviously :) and not all in one big drop. I'll close this PR for now so we don't continue to get spammed by new pushes, and create a chain of branches in my fork so you can pull from what you wish.

The current goal I have in my fork, just to make that clear, is to get to the point where I can have a storage module that uses ecto for persistence and an erlang process for update notification.

Cheers!

tompave:
👍

PR #5, opened 2017-05-06
Centralize serialization code

aseigo:
Centralizes the serialization code to/from Redis. This will make it easier in future to create alternative storage implementations, and in the more immediate it makes the code a bit tidier.

(Note: this is the same as a previous pull request which got 'poluted' with additional changes. I've broken those out into a series of branches in my fork, and will continue moving forward one branch at a time towards my own personal goal here, which is an Ecto storage + BEAM process for change notification persistence layer. This way upstream can pick up changes as far it makes sense to do so.)

### DIFF
--- a/lib/fun_with_flags/flag.ex
+++ b/lib/fun_with_flags/flag.ex
@@ -13,16 +13,6 @@ defmodule FunWithFlags.Flag do
   end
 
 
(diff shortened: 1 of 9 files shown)


tompave:
Thanks for this 👍

PR #6, opened 2017-05-09
Feature/store flexibility ps02

aseigo:
Next potential step in tidying the storage part: drop the storage supervisor, and have the storage implementation take care of all of its own children (granting flexibility on how notifications are serviced, for instance).

The rational for this change is that the main supervisor was already doing part of the storage-specific child process setup (for notifications), while the storage supervisor was handling the rest. My thought was that this would be cleaner if done all in one place. It makes the code a bit easier to reason about, mostly by isolating the storage modules, config and supervisor tree a bit better from each other.

That said, this is one of two directions this could be done in. One is as it is done here, with just a single supervisor for the library. The other is to keep the storage supervisor and move all the storage related children (which includes change notifications) into that, so that the supervision tree would look like:

  Application Supervisor -> Storage Supervisor -> Storage Related Processes

In future, if other processes are brought into the library that need supervision, it could be cleaner to do the latter. However, I felt that this additional bit of complexity was not justified at this time. ("Don't make things more flexible than necessary" principle) Should a more complex process tree be needed later on, putting the storage related processes under their own supervisor would be quite easy, so it is safe to delay that move (if it is ever needed).


tompave:
Hi, thank you for taking the time to submit this.

While I can in theory agree with simplifying the supervisors, this PR also changes the internal API that supports different adapters, so I don't think I can accept it.

Also, unfortunately, I already have plan on how this should evolve, and I simply do not have the time to work on them right now.

aseigo:
>  this PR also changes the internal API that supports different adapters, so I don't think I can accept it.

No problem :) On that note, however: the internal API change moves the responsibility for internal storage details (does it support notifications? if so, what is the module for that?) away from Config module and puts it in the hands of the storage module. 

This is, at least imho, breaking encapsulation of the storage module. The functions in question were just convenience wrappers, but ended up replacing/obscuring the actual API being used. As you evolve this forward, hopefully these things will end up naturally moving  into their respective relevant modules.

Issue #14, opened 2017-12-13
Should we be compiling in configs


olafura:
This line means that umbrella apps won't work unless they have the application where the config was initially defined as a dependency:
https://github.com/tompave/fun_with_flags/blob/3aa0d11cd050e9690d7b4d551674355f6dbcd03d/lib/fun_with_flags/store/persistent/ecto.ex#L12

tompave:
Hi, thank you for flagging the issue.

I'm not sure I follow, sorry. I'm not too familiar with umbrella apps, and I've never tries to use this package in an umbrella app.

Some issues with those module attributes are mentioned [in the readme](https://github.com/tompave/fun_with_flags#configuration-changes-have-no-effect-in-mix_envdev), but it seems that you're reporting something different.

Can you please add more details? If you can propose a solution that keeps the tests green, I'd be happy to consider it.

olafura:
Hope the PR explains the problem better than my description in the issue above ;)

tompave:
Regarding the title question:

> Should we be compiling in configs?

I initially did it because I didn't want the performance bottleneck of reading those values from an ETS table all the time, where they would end up if they were simply read dynamically from Mix config.
I guess it's acceptable for things that are not read very often (the Ecto repo is a good example), but I'm not sure about this, for example:
https://github.com/tompave/fun_with_flags/blob/master/lib/fun_with_flags.ex#L30

olafura:
@tompave sure I trust your call, have you benchmarked this ETS are usually really performant? If it's really a problem there are some macro possibilities which would allow you to recompile on change.

olafura:
@tompave I'm not seeing any difference with `:timer.tc(FunWithFlags, :all_flag_names, [])` between master and the PR. You just tell me if you me to revert any changes in the PR, it's pretty easy todo.

tompave:
I've done some synthetic benchmarks, but they might not be too relevant. Rather than speed, I was mainly concerned with the number of inter-process messages and copying data between processes, as that's what happens when you read from an ETS table.

I've been meaning to revisit those decisions though. I considered [fastglobal](https://hex.pm/packages/fastglobal), but now I realize that it would have the same issue that you reported here.

tompave:
> I'm not seeing any difference with :timer.tc(FunWithFlags, :all_flag_names, []) between master and the PR. You just tell me if you me to revert any changes in the PR, it's pretty easy todo.

[...]

PR #15, opened 2017-12-13
Fix compiled in configs for working with umbrella

olafura:
The problem with compiling in configs is that if you have
umbrella projects that are both setting the same config,
either one will get their compiled config in and the other
will fail. Since you can't introduce circular deps in your
umbrella application you would have to introduce an other
application that all of the apps that need FunWithFlags
would need to depend on.

Fixes: #14

### DIFF
--- a/lib/fun_with_flags/store/persistent.ex
+++ b/lib/fun_with_flags/store/persistent.ex
@@ -1,5 +1,0 @@
-defmodule FunWithFlags.Store.Persistent do
-  @moduledoc false
-
-  def adapter, do: FunWithFlags.Config.persistence_adapter
-end
--- a/lib/fun_with_flags/store/persistent.ex
+++ b/lib/fun_with_flags/store/persistent.ex
@@ -1,7 +1,5 @@
 defmodule FunWithFlags.Store.Persistent do
   @moduledoc false
 
-  @adapter FunWithFlags.Config.persistence_adapter
-
-  def adapter, do: @adapter
+  def adapter, do: FunWithFlags.Config.persistence_adapter
 end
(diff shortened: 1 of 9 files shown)


tompave:
Hi, thank you for the PR.
The Travis-CI build didn't run because of a misconfiguration (it was enabled for branches, not PRs). Can you please trigger it now? (you can do so with `git commit --amend && git push -f`)

olafura:
@tompave No problem this should hopefully work ;)

tompave:
Thank you for this.

All tests are green, so this would be good to go.
I just want to check a couple of things first:

1) reproduce the problem with an umbrella app.
2) run some benchmarks. Things like `@repo` and `@persistence` can easily be made dynamically because they're not called often, but doing the same with `@store` could cause a performance regression. I just want to properly assess the impact.

tompave:
See the original issue for context: https://github.com/tompave/fun_with_flags/issues/14#issuecomment-351454286

In terms of benchmarks, I'm thinking of concurrent calls from thousands of processes (e.g. cowboy processes in a Phoenix app), as that's where the performance hit would come from, I think.

lib/fun_with_flags/store/persistent.ex
```diff
@@ -1,7 +1,5 @@
 defmodule FunWithFlags.Store.Persistent do
   @moduledoc false
 
-  @adapter FunWithFlags.Config.persistence_adapter
-
-  def adapter, do: @adapter
+  def adapter, do: FunWithFlags.Config.persistence_adapter
```
tompave: The `@adapter` attribute was placed here just because it was the right scope, semantically speaking. Since we're making this dynamic, we can avoid a chain of function calls by removing this and having the callers use `Config.persistence_adapter` directly.

olafura: Cool

tompave:
I'm closing this because of what I explained in https://github.com/tompave/fun_with_flags/issues/14#issuecomment-355246043

PR #16, opened 2018-01-11
Update README for redis persistence

Christopher-Bui:
localhost was specified with single quotes and not double quotes, leading to an error if blindly copying and pasting.


tompave:
Ops. Thanks!

Issue #18, opened 2018-02-05
"Looks like ... Redis" when using ecto


coryodaniel:
This isn't a bug, but I did fumble through getting fun_with_flags set up for a while. I turns out I wasn't planning on using caching because I was working in dev/test environment trying to set up fun_with_flags with Ecto. I kept getting an error message despite setting Ecto as my adapter saying that I was missing a Redis dependency.

It turns out this error was specifically because I had neglected to turn caching off and it defaults to Redis. 

Do you think this is better addressed in the README or modifying the error message to reflect that you can also simply disable caching.

coryodaniel:
https://github.com/tompave/fun_with_flags/pull/20

PR #20, opened 2018-02-05
Clarifying error regarding missing Redis dep

coryodaniel:
Currently when you are using Ecto as the persistence store you will
receiving an error message that Redis is a dependency if you have
caching enabled. I updated the error message to reflect that caching
can also be disabled.

### DIFF
--- a/priv/ecto_repo/migrations/00000000000000_create_feature_flags_table.exs
+++ b/priv/ecto_repo/migrations/00000000000000_create_feature_flags_table.exs
@@ -2,7 +2,8 @@ defmodule FunWithFlags.Dev.EctoRepo.Migrations.CreateFeatureFlagsTable do
   use Ecto.Migration
 
   def up do
-    create table(:fun_with_flags_toggles) do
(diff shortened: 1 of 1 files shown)


tompave:
Closes https://github.com/tompave/fun_with_flags/issues/18

tompave:
Hi! Thank you for trying the library and for the PR.

A couple of comments:

1. Looks like your creating [two PRs](https://github.com/tompave/fun_with_flags/pull/21) from the same master branch in your fork, and both PRs try to apply the same changes. Can you please separate them to keep things tidy?
2. As you mentioned in the [accompanying issue](https://github.com/tompave/fun_with_flags/issues/18), it would be good to add clearer instructions to the Readme too.
3. Rather than disabling the cache, it would be better to just disable the change notifications. You can for example use caching without notifications. See https://github.com/tompave/fun_with_flags/search?utf8=%E2%9C%93&q=cache_bust_notifications

coryodaniel:
In favor of: https://github.com/tompave/fun_with_flags/pull/22

Issue #19, opened 2018-02-05
null value in column "id" violates not-null constraint


coryodaniel:
I was getting an error message when enabling a flag from postgrex: 
```
** (Postgrex.Error) ERROR 23502 (not_null_violation): null value in column "id" violates not-null constraint
```

In my application I have Ecto configured to use UUIDs by default. You could add some logic to the `Record` field to figure this out probably, although, it may be simpler to modify the migration for the table to explicitly set the primary key to an `serial` type:

```elixir
create table(:fun_with_flags_toggles, primary_key: false) do
  add :id, :serial, primary_key: true
  add :flag_name, :string
  add :gate_type, :string
  add :target, :string
  add :enabled, :boolean
end
```

coryodaniel:
https://github.com/tompave/fun_with_flags/pull/23

PR #21, opened 2018-02-05
Explicitly set primary key to serial type

coryodaniel:
Explicitly set PK to type SERIAL for users that have their default Ecto configuration to use UUIDs.

### DIFF
--- a/priv/ecto_repo/migrations/00000000000000_create_feature_flags_table.exs
+++ b/priv/ecto_repo/migrations/00000000000000_create_feature_flags_table.exs
@@ -2,7 +2,8 @@ defmodule FunWithFlags.Dev.EctoRepo.Migrations.CreateFeatureFlagsTable do
   use Ecto.Migration
 
   def up do
-    create table(:fun_with_flags_toggles) do
+    create table(:fun_with_flags_toggles, primary_key: false) do
+      add :id, :serial, primary_key: true
       add :flag_name, :string
       add :gate_type, :string
       add :target, :string


tompave:
Closes https://github.com/tompave/fun_with_flags/issues/19

tompave:
Hi! Thank you for trying the library and for the PR.

A couple of comments:

1. Looks like your creating [two PRs](https://github.com/tompave/fun_with_flags/pull/20) from the same master branch in your fork, and both PRs try to apply the same changes. Can you please separate them to keep things tidy?
2. So just changing the migration like that fixes the problem? Isn't `bigserial` [the default](https://hexdocs.pm/ecto/2.2.8/Ecto.Migration.html#module-repo-configuration)? (In postgres, [a bigserial is just like a serial](https://www.postgresql.org/docs/9.6/static/datatype-numeric.html) but with a wider range). I'm asking because I would assume that explicitly setting the primary key to serial (instead of bigserial) would not change the table too much (the primary key would still be an integer).
3. Can you provide an example of your Ecto configuration, please? So that I can reproduce the issue locally. Ideally I would prefer a solution that works with any set up, rather than changing the migration.
4. How would this work with Ecto adapters other than postgrex?
5. How about providing an alternative migration instead (e.g. commented out), that people can use instead if their primary keys are UUIDs? The reason is that I'm not too keen to just change the default migration, as it would mean that new code added in the future would only be tested with the new one.

tompave:
If I get this right, you're setting the PK to `serial` so that Ecto knows that this table has an integer PK, as opposed to all the other tables in your DB that have a UUID PK. Ok, can we make it a `bigserial` instead, since that's the deafult?

How does that lead to this error?
```
** (Postgrex.Error) ERROR 23502 (not_null_violation): null value in column "id" violates not-null constraint
```

coryodaniel:
Bigserial would work. It gets that error because Ecto generates the UUID not PG, but PG creates the serial type.  Ill update the PRs. Meant to send from a branch.

coryodaniel:
Actually, after looking at it some more, why even have a surrogate primary key? It appears there is a natural primary key `[:flag_name, :gate_type, :target]`


The conf that causes this is:

```elixir
config :my_app, ecto_repos: [MyApp.Repo]
config :my_app, MyApp.Repo,
  migration_primary_key: [type: :binary_id]
```

coryodaniel:
In favor of: https://github.com/tompave/fun_with_flags/pull/23

PR #23, opened 2018-02-08
Changed ecto migration to use a natural PK

### LINKED ISSUE #19: null value in column "id" violates not-null constraint

coryodaniel:
If a project sets ecto's `migration_primary_key` to binary_id, the
FunWithFlags Record module fails to insert the primary key because ecto
normally autogenerates this, but `@primary_key {:id, :binary_id,
autogenerate: true}` has to be set on the module.

Closes #19

### DIFF
--- a/lib/fun_with_flags/store/persistent/ecto/record.ex
+++ b/lib/fun_with_flags/store/persistent/ecto/record.ex
@@ -5,6 +5,7 @@ defmodule FunWithFlags.Store.Persistent.Ecto.Record do
   use Ecto.Schema
   import Ecto.Changeset
 
+  @primary_key false
   schema "fun_with_flags_toggles" do
     field :flag_name, :string, primary_key: true
     field :gate_type, :string, primary_key: true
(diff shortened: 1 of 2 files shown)


tompave:
Ops, looks like the changes got mangled when you split https://github.com/tompave/fun_with_flags/pull/21

tompave:
Thanks! LGTM

PR #29, opened 2018-06-01
Allow configuration of redis using {:system, envvar}

seangeo:
Hi,

Thanks for this app, really enjoying using it. Just came across one quirk when deploying to Heroku.

When deploying to Heroku, you need to set the Redis url using an environment variable. If you do this in the config like this:

```elixir
config :fun_with_flags, :redis, System.get_env("REDIS_URL")
```

and change the value in the environment config, it won't be reflected in the app until a redeploy forces a recompile. So instead it's common practise to use a system tuple like so:

```elixir
config :fun_with_flags, :redis, {:system, "REDIS_URL"}
```

This PR adds support for this configuration option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -512,6 +512,9 @@ config :fun_with_flags, :redis,
 
 # a URL string can be used instead
 config :fun_with_flags, :redis, "redis://locahost:6379/0"
+
+# a {:system, name} tuple can be used to read from the environment
+config :fun_with_flags, :redis, {:system, "REDIS_URL"}
 ```
 
 ### Persistence Adapters
--- a/lib/fun_with_flags/config.ex
+++ b/lib/fun_with_flags/config.ex
@@ -27,6 +27,8 @@ defmodule FunWithFlags.Config do
         uri
       opts when is_list(opts) ->
         Keyword.merge(@default_redis_config, opts)
+      {:system, var} when is_binary(var) ->
+        System.get_env(var)
     end
   end
 
--- a/test/fun_with_flags/config_test.exs
+++ b/test/fun_with_flags/config_test.exs
@@ -28,6 +28,12 @@ defmodule FunWithFlags.ConfigTest do
     assert            2000 == Config.redis_config[:port]
     assert              42 == Config.redis_config[:database]
 
+    # Whe configured with a {:system, env} tuple it looks up the value in the env
+    System.put_env("123_TEST_REDIS_URL", url)
+    configure_redis_with({:system, "123_TEST_REDIS_URL"})
+    assert url == Config.redis_config
+    System.delete_env("123_TEST_REDIS_URL")
+
     # cleanup
     configure_redis_with(defaults)
   end


tompave:
Hello,

thank you for the PR and for using the library. :-)

---

I see what you mean and that's a normal issue with Elixir and Mix configuration, and not just limited to Heroku. The Mix config files are in fact only parsed and executed at compile time, not when the application boots, and this has historically caused a lot of confusion.

Judging by what you described, I assume that you're using a custom Heroku buildpack that compiles the Elixir application into an Erlang release when you push new code and the slug is built. Heroku will compile a slug when you push new code, but not when you change the env. Changing an env variable will simply cause Heroku to reboot the dynos with the same slug, which causes the problem you've described because the previous env value is now part of the compiled Erlang bytecode.

The exact same thing happens whenever you re-deploy an already compiled Elixir app (i.e. an Erlang release) with a different config in the OS environment. This includes "old school" releases and any modern flavour of containers.

---

Back to the library and the problem you've encountered, the config snippets in the readme are just simple examples to show what's expected to go where, e.g. "this is supposed to be a URL provided as a binary", but as you've discovered they can be inflexible if used in production code without changes.

While your solution works, I am afraid that I cannot accept the PR because at the moment things are a bit uncertain around configuration, and it seems that the situation will evolve soon.
See [this proposal on the Elixir Forum](https://elixirforum.com/t/proposal-moving-towards-discoverable-config-files/14302) for example, or [this blog post](http://www.theerlangelist.com/article/rethinking_app_env) that describes the current problems. It's also worth pointing out that the community guidelines have evolved since I started working on this library (my first serious Elixir project!), and [today's recommendation is to not use the Application environment at all](https://hexdocs.pm/elixir/master/library-guidelines.html#avoid-application-configuration). With that in mind, I'd rather not include ad-hoc solutions that don't go in the same direction as the community. Sorry about that :-(

I can propose some approaches to get you unblocked though :-)

In similar situation, I've used [`deferred_config`](https://hex.pm/packages/deferred_config) to solve this problem, and I've found it a flexible and low-touch solution that just hooks into the current configuration set up. It will allow you to use the `:system` tuples with any Application env configuration.

Also, if your Heroku buildpack uses Distillery to compile the app, you can try to use the [replace ENV vars](https://hexdocs.pm/distillery/getting-started.html#vm-configuration) functionality to inject the dynamic env into the application when the dynos start.

I've also heard of folks running Elixir applications on Heroku with Mix. For example, with a Phoenix app:

```
MIX_ENV=prod mix phx.server
```
That would load the Mix config files at boot time, parse the OS environment dynamically, and make your Elixir application behave more like Ruby on Rails, Django or Node.js. Please keep in mind that I have **not** used this approach personally, some folks do not recommend it, and that I've mentioned it just for completeness.

seangeo:
Hi 

Thanks for the detailed reply. I do understand that the community is very much in flux around config at the moment and not using application config at all would be a good option. However part of this is on libraries as well. It appears to me that fun_with_flags provides no other option but to configure the redis url though the application config and with quite strict pattern matching. Unless I’m missing something?

So I guess the question is, if you won’t accept this solution is there a plan to include a more general way to configure the redis url at runtime without having to use additional libraries or is the plan to wait and see what happens with the current discussions around config? 

That said, thanks very much for the recommendations, deferred_config does look like a good work around to the whole problem, one I hadn’t seen before. It does looks like deferred config needs to be called before the application, fun_with_flags in this case, is started. So to use it, I’ll have to mess with application start. 

Thanks again for the pointers and the library. :)

tompave:
Hi,

I'm glad it was of some help. :-)

> So I guess the question is, if you won’t accept this solution is there a plan to include a more general way to configure the redis url at runtime without having to use additional libraries or is the plan to wait and see what happens with the current discussions around config?

Both are correct.

I'm waiting to see what happens, and especially if the idea of dynamically loading Mix/App config files at runtime gets implemented.

The other thing I've considered is to use the `init` callback approach that is recommended today, so that you would `use FunWithFlags` into your custom module, provide an `init` where you load the config in any way you want, and then start your custom module as part of the application's supervision tree. It's not a change I can do without releasing a `2.0` though, and before committing to any change of that scope I'd really rather see what happens with the language.

seangeo:
Sounds reasonable to me. Thanks again. :)

seangeo:
Hi @tompave thanks for that!

PR #34, opened 2019-02-18
Fix compiler warning and add --warnings-as-errors to CI

asummers:
```
warning: invalid type annotation. When using the | operator to represent the union of types, make sure to wrap type annotations in parentheses: FunWithFlags :: Gate.t()
  lib/fun_with_flags/store/serializer/redis.ex:8
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,9 @@ otp_release:
 env:
   - CACHE_ENABLED=true  TEST_OPTS='--exclude phoenix_pubsub --exclude ecto_persistence'
   - CACHE_ENABLED=false TEST_OPTS='--only integration'
-  - CACHE_ENABLED=true  PUBSUB_BROKER=phoenix_pubsub TEST_OPTS='--no-start --exclude redis_pubsub --exclude ecto_persistence --exclude phoenix_pubsub:with_ecto --include phoenix_pubsub:with_redis --include phoenix_pubsub:true' 
+  - CACHE_ENABLED=true  PUBSUB_BROKER=phoenix_pubsub TEST_OPTS='--no-start --exclude redis_pubsub --exclude ecto_persistence --exclude phoenix_pubsub:with_ecto --include phoenix_pubsub:with_redis --include phoenix_pubsub:true'
   - CACHE_ENABLED=false PUBSUB_BROKER=phoenix_pubsub TEST_OPTS='--no-start --only integration'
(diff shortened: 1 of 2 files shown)


lib/fun_with_flags/store/serializer/redis.ex
```diff
@@ -5,7 +5,7 @@ defmodule FunWithFlags.Store.Serializer.Redis do
 
   @type redis_hash_pair :: [String.t]
 
-  @spec serialize(FunWithFlags::Gate.t) :: redis_hash_pair
+  @spec serialize(fun_with_flags :: Gate.t()) :: redis_hash_pair
```
tompave: Ah, good catch!

Just one thing. This is actually me accidentally using Ruby's module navigation operator. I wrote `FunWithFlags::Gate` but I actually meant `FunWithFlags.Gate`. So this just fixes the compiler warning, but actually breaks the typespec.

Can you please just change it to `FunWithFlags.Gate.t`?

tompave:
Hi, thank you for taking care of this! ❤️

Once you address that comment, I'll be happy to merge this.

asummers:
Oops shame on me for not looking at the types.

.travis.yml
```diff
@@ -40,6 +40,7 @@ services:
 addons:
   postgresql: "9.6"
 script:
+  - mix compile --warnings-as-errors
```
tompave: Will this also fail with deprecation warnings? For example if `mix deps.get` pulls in an updated dependency and the compiler raises a warning because of a deprecated method?

I know that it doesn't break if the deprecation warning comes _from a dependency_, but I'm thinking about deprecation warnings coming from FunWithFlags itself.

asummers:
It would only break if a deprecated call is added in any of the versions of the build matrix. So if you added a HashDict or something like that. According to Travis there are currently no deprecated calls so they would have to be added via future Elixir versions or explicitly added to the current code. The usual way around that problem is to do a ELIXIR_VERSION != “x.y.z” && mix compile —warnings-as-errors type thing. But that’s a problem your code does not currently have.

asummers:
The other way around that problem is detecting elixir version in the code and bifurcating at compile time via macros to the old/new call. That’s what Elixir does internally

lib/fun_with_flags/store/serializer/redis.ex
```diff
@@ -5,7 +5,7 @@ defmodule FunWithFlags.Store.Serializer.Redis do
 
   @type redis_hash_pair :: [String.t]
 
-  @spec serialize(FunWithFlags::Gate.t) :: redis_hash_pair
+  @spec serialize(FunWithFlags.Gate.t()) :: redis_hash_pair
```
tompave: Hi, sorry, can you remove the parenthesis from this, so that it's just `FunWithFlags.Gate.t`. That's the style I've used in the all the typespecs in this codebase.

tompave:
Thanks for this :-)

tompave:
Fix released in version `1.2.1` https://hex.pm/packages/fun_with_flags/1.2.1
Thanks @asummers!

Issue #40, opened 2019-03-26
'conflict_target' in Ecto persistence prevents use with MySQL


stewart:
Howdy! I'm attempting to integrate `fun_with_flags` into a Phoenix app that's using a MySQL database. Unfortunately, at the moment it doesn't quite behave as expected:

```
iex(1)> FunWithFlags.enable(:test_flag)
** (ArgumentError) The :conflict_target option is not supported in insert/insert_all by MySQL
    (ecto_sql) lib/ecto/adapters/mysql/connection.ex:972: Ecto.Adapters.MySQL.Connection.error!/2
    (ecto_sql) lib/ecto/adapters/mysql/connection.ex:149: Ecto.Adapters.MySQL.Connection.insert/6
    (ecto_sql) lib/ecto/adapters/mysql.ex:208: Ecto.Adapters.MySQL.insert/6
    (ecto) lib/ecto/repo/schema.ex:651: Ecto.Repo.Schema.apply/4
    (ecto) lib/ecto/repo/schema.ex:264: anonymous fn/15 in Ecto.Repo.Schema.do_insert/3
    (fun_with_flags) lib/fun_with_flags/store/persistent/ecto.ex:193: FunWithFlags.Store.Persistent.Ecto.do_insert/3
    (fun_with_flags) lib/fun_with_flags/store/persistent/ecto.ex:71: FunWithFlags.Store.Persistent.Ecto.put/2
    (fun_with_flags) lib/fun_with_flags/store.ex:36: FunWithFlags.Store.put/2
    (fun_with_flags) lib/fun_with_flags.ex:185: FunWithFlags.enable/2
```

It should be possible to add this option to the insert statement dynamically, based on the result of `Repo.__adapter__` (AFAIK, MySQL is the special case here) - but I'm not certain if that's the best way to approach this.

If you'd welcome a patch implementing the above suggestion, let me know.

tompave:
Hello! 👋 

Thank you for trying the library and for raising the issue.

I see. Unfortunately I've developed using Postgres locally, and I've never tested it with MySQL. I just assumed that Ecto's API and its options would work with any RDBMS supported by the `ecto` package (and, today, the `ecto_sql` package).

It seems that that was naive on my part.

> It should be possible to add this option to the insert statement dynamically, based on the result of Repo.__adapter__ (AFAIK, MySQL is the special case here)

[...]

PR #41, opened 2019-03-27
Ecto Persistence: Customize UPSERT options for MySQL adapter

stewart:
`ecto_sql` [supports UPSERT operations](https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert/2-upserts) for both the MySQL and Postgres adapters. Postgres requires the `conflict_target` option be supplied when performing updates, while MySQL does not support it at all.

To allow supporting both of these backends, we'll use the Repo's adapter to differentiate what kind of database we're talking to, and use that to generate the appropriate insert options.

This addresses #40. As mentioned in that issue, `.travis.yml` changes are required to begin testing against both MySQL and Postgres backends - I wasn't fully comfortable making those changes myself, as the existing build matrix and test setup is reasonably complex.

### DIFF
--- a/lib/fun_with_flags/store/persistent/ecto.ex
+++ b/lib/fun_with_flags/store/persistent/ecto.ex
@@ -63,10 +63,7 @@ defmodule FunWithFlags.Store.Persistent.Ecto do
 
   def put(flag_name, gate = %Gate{}) do
     changeset = Record.build(flag_name, gate)
-    options = [
-      on_conflict: [set: [enabled: gate.enabled]],
-      conflict_target: [:flag_name, :gate_type, :target] # the unique index
-    ]
+    options = upsert_options(gate)
 
     case do_insert(flag_name, changeset, options) do
       {:ok, flag} ->
@@ -185,6 +182,21 @@ defmodule FunWithFlags.Store.Persistent.Ecto do
       @repo,
       "LOCK TABLE fun_with_flags_toggles IN SHARE ROW EXCLUSIVE MODE;"
     )
+  end
+
+
+  defp upsert_options(gate = %Gate{}) do
+    options = [on_conflict: [set: [enabled: gate.enabled]]]
+
+    case @repo.__adapter__ do
+      # MySQL's UPSERT support doesn't allow a specific conflict_target
+      Ecto.Adapters.MySQL ->
+        options
+
+      # For other databases, specify the unique index
+      _ ->
+        options ++ [conflict_target: [:flag_name, :gate_type, :target]]
+    end
   end
 
 
--- a/lib/fun_with_flags/store/persistent/ecto/null_repo.ex
+++ b/lib/fun_with_flags/store/persistent/ecto/null_repo.ex
@@ -12,4 +12,5 @@ defmodule FunWithFlags.NullEctoRepo do
   def update(_, _),    do: raise(@error_msg)
   def delete_all(_),   do: raise(@error_msg)
   def transaction(_),  do: raise(@error_msg)
+  def __adapter__(),   do: raise(@error_msg)
 end


tompave:
Hi, thanks for the PR :-)

All the non-ecto build tasks are failing with this:

```
warning: function FunWithFlags.NullEctoRepo.__adapter__/0 is undefined or private
  lib/fun_with_flags/store/persistent/ecto.ex:191
The command "mix compile --warnings-as-errors" exited with 1.
```

You'll need to add a stubbed function to `NullEctoRepo`.

stewart:
Apologies for the delayed feedback loop, [Travis has been having some issues today][issues]. The updated build's just finished, with a [couple][1] [failures][2] that don't appear to be related to this patch - both of the jobs are using the Redis persistence adapter.

It's possible these are one-off failures related to Travis' partial outage, but I can't ask Travis to retry the build without making an additional commit - could you hit the Retry button when you have a moment?

[issues]: https://www.traviscistatus.com/incidents/my5wm56npf7q
[1]: https://travis-ci.org/tompave/fun_with_flags/jobs/512586615
[2]: https://travis-ci.org/tompave/fun_with_flags/jobs/512586641

tompave:
Yup, the looked like flakeys. I've restarted them and now everything is green.

This seems to work 👍 
Have you tried this branch in your project using MySQL?

I'll tweak the Travis matrix after this is merged.

stewart:
Yep, it's behaving as expected with this change.

tompave:
Nice. Thank you for working on this. 👍 
I need to find some time to test it locally first, but it looks like this is good to be merged.

Before releasing a new version, I want to find a way to get the Ecto test to run with both Postgres and MySQL in CI.

tompave:
Testing it locally with MySQL `8.0.15` and a Phoenix app using https://github.com/tompave/fun_with_flags_ui.

When I try to add a percentage gate I get this error, reported through Phoenix:

```
(1064): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'IN SHARE ROW EXCLUSIVE MODE' at line 1
```

tompave:
What version of MySQL are you using?

stewart:
I'm not actually certain, or able to easily tell - this app's hosting situation is somewhat opaque to the development team, for multiple reasons (part of the reason we need feature flags!).

Seems like that's an issue with the in-transaction table lock for the percentage gate `put` operation. I'm not immediately certain what a direct translation of that to MySQL looks like (I'm more used to Postgres, personally), but am actually concerned it could lead to a deadlock -at least in MySQL, explicit locks are not transaction-safe.

We're not planning to use the percentage flag functionality anytime soon, so perhaps that can be filed as a separate issue and a fix devised there?

tompave:
> I'm not actually certain, or able to easily tell - this app's hosting situation is somewhat opaque to the development team, for multiple reasons.

I see. You can try with this:

```sql
SHOW VARIABLES LIKE "%version%";
```

If you don't have access to the `mysql` client, you can run it with `iex -S mix` in your Phoenix app:

```elixir
Ecto.Adapters.SQL.query!(YourApp.Repo, "SHOW VARIABLES LIKE "%version%";")
```

tompave:
> Seems like that's an issue with the in-transaction table lock for the percentage gate put operation. 

Yes, it's definitely an issue with this:
https://github.com/tompave/fun_with_flags/blob/v1.2.1/lib/fun_with_flags/store/persistent/ecto.ex#L183-L188

The table lock is required because the unique three-column index alone wouldn't prevent duplicated rows for the percentage gates. A different table design would make the table lock not necessary, but when I added the percentage gates I was mainly concerned with keeping it a small and retro-compatible version update. Perhaps it's something I could revisit for version 2.0.

> I'm not immediately certain what a direct translation of that to MySQL looks like (I'm more used to Postgres, personally), but am actually concerned it could lead to a deadlock -at least in MySQL, explicit locks are not transaction-safe.

I'm more a Postgres person too. I've looked at the [MySQL docs for table locking](https://dev.mysql.com/doc/refman/8.0/en/lock-tables.html#lock-tables-and-transactions), and it's not really clear to me yet what the right pattern is. The fact that locks are not released when a transaction commits or rolls back is concerning.

> We're not planning to use the percentage flag functionality anytime soon, so perhaps that can be filed as a separate issue and a fix devised there?

I see your point, but in that case I suggest you use your fork instead of getting this partial fix merged. Since with this alone I can't really claim that "it now works with MySQL", I wouldn't release a new version for it -- I wouldn't even keep it on the master branch, but on a `mysql_fixes` feature branch. So even if I merged this PR, you'd still have to use a git source for the package declaration in your Mixfile, and not pull it from Hex.

The reason is that this is the first time I hear about this problem. I suspect that the reason is most people are using this package with either Redis or Ecto+Postgres; if anyone else tried to use it with MySQL in the past and got blocked because of the problem you encountered, they didn't bother reporting it here on GitHub.

At the moment the situation seems to be that most users of the package can use it fine with their set up, and I could just add to the docs that the Ecto adapter works with Postgres only, even though it's not ideal. On the other hand, if I were to merge this as it is, I wouldn't really consider it ready for release. The changelog entry would be: "now it partially works with MySQL, but not everything". 🤔 

So I'd rather either fix everything or nothing at all.

I'll now create a new `mysql_fixes` branch that you could use as target of this PR, where I'll also add support to run the test suite with MySQL. I guess we can merge this there, and then use that as target for any new work on MySQL compatibility.

tompave:
I've amended the docs: https://github.com/tompave/fun_with_flags/commit/a2e8e443faf5757da97a114bea1f638e520c64f3

And I've created the [`mysql_support`](https://github.com/tompave/fun_with_flags/tree/mysql_support) branch, that you can target with this PR to get it merged.

tompave:
Update: I've given a stub at running the tests on MySQL. Currently a bit blocked locally because of this: https://github.com/xerions/mariaex/issues/222

That issue also seems to suggest that `mariaex` is not compatible with MySQL 8.0. 🤔

stewart:
Thanks for the README update. Based on the other discoveries in this thread, I think it's reasonable to use a `mysql_support` branch if this is a configuration you're interested in supporting in the future, and I've updated this PR to target it, as requested.

I am reasonably confident we're using a `5.7.20+` release of MySQL, but will verify for certain when I'm able.

tompave:
Ok, I've had to downgrade my local env from MySQL 8.0 to 5.7, and now Mariaex can connect.
I've already done some work in the `mysql_support` branch to make it work with MySQL. If you want to check it out, you can use `bin/console_ecto mysql` to start a `iex` console connected to MySQL.

I'll now merge this and add some travis tasks for MySQL specifically.

PR #42, opened 2019-04-02
MySQL support

tompave:
This branch tracks the progress to add support for MySQL to the Ecto persistence adapter, which currently only works with PostgreSQL.

The current status is that this branch contains some changes to work with MySQL on local dev and test, as well as on CI.

~The builds still fail with MySQL because of [how the table is locked](https://github.com/tompave/fun_with_flags/blob/v1.2.1/lib/fun_with_flags/store/persistent/ecto.ex#L183-L188) using the Postgres syntax `IN SHARE ROW EXCLUSIVE MODE`, which MySQL doesn't understand. MySQL also seems to have different guarantees in terms of table locks within transactions ([docs](https://dev.mysql.com/doc/refman/8.0/en/lock-tables.html#lock-tables-and-transactions)), which makes adapting the logic not straightforward.~ (solved)

The table lock is required because the unique three-column index alone wouldn't prevent duplicated rows for the percentage gates. A different table design would make the table lock not necessary, but when I added the percentage gates I was mainly concerned with keeping it a small and retro-compatible version update. Perhaps it's something I could revisit for version 2.0, but I don't have any concrete plan at the moment.

Thank you @stewart for starting this work by raising issue https://github.com/tompave/fun_with_flags/issues/40 and working on the first fix, with PR https://github.com/tompave/fun_with_flags/pull/41.


tompave:
Hi @stewart, I think that this is now working. Do you want to give it a spin?

stewart:
Thanks @tompave -  this changeset works as expected in my development environment, using `FunWithFlags.UI` and the console to test creating/updating various types of flags.

I'll be able to confirm more concretely once we've got it deployed to a staging environment for further testing, but I don't see any immediate problems.

tompave:
Ok. Leaving this here for the moment till you can confirm it works in a deployed env.

tompave:
Hi @stewart, any update on this?

stewart:
I apologize, but we're still waiting for the go-ahead for a staging deploy, which should happen tomorrow. In local development, my team has not seen any issues with this branch.

I also noted the [most recent release](https://github.com/elixir-ecto/ecto_sql/blob/master/CHANGELOG.md#v310-2019-04-02) of `ecto_sql` includes a new MySQL adapter (as they're looking to stop using `mariaex` in the future), `Ecto.Adapters.MyXQL`. It's likely worth adding this to the check in `db_type/0`.

stewart:
Still haven't gotten a green-light for our planned staging deploy, unfortunately - it's currently planned for tomorrow.

In the interest of keeping this PR moving, I've manually verified the MySQL persistence strategy for both the boolean and percentage-gate flags works as expected on both MySQL 5.7.25 and MariaDB 10.3.14, using both the `mariaex` and `myxql` Ecto adapters. Behaviour under load was tested by spawning a swarm of 100-500 Tasks that all attempted to set the same flag, and I didn't experience any deadlocks in this testing.

I additionally attempted to test with MySQL 8.0 and `myxql`, but was blocked by authorization issues that would require additional SSL plumbing to correct:

```
14:34:58.358 [error] GenServer #PID<0.312.0> terminating
** (MyXQL.Error) (2061) (CR_AUTH_PLUGIN_ERR) Authentication plugin 'caching_sha2_password' reported error: Authentication requires secure connection
```

Everything looks good to me.

tompave:
Thanks for running those tests. 👍

stewart:
Can also confirm this branch is behaving as expected in our staging environments. Thank you for the additional MySQL support changes, and your patience with our slow turn-around for testing them.

tompave:
Cool, thanks for confirming it. 👍 

This is now on master. Before releasing a new version, I want to add some more documentation and changelog entries, and I want to verify that everything still works with both `ecto` 3.0 and 3.1, and with both `mariaex` and `myxql`.

tompave:
Ok, released on Hex.pm as v1.3.0

PR #43, opened 2019-04-08
fix typo in README

chad-g-adams:


tompave:
Hi, thank you for this.

For some reason the CI check is still pending (yellow) even though the build has completed successfully. I'll restart it to see if it unblocks it.

Issue #47, opened 2019-06-26
Improved runtime config support


bernardd:
Hi - I want to use this library in our app, however because we run it in multiple different environments from a single release, all our config comes in through environment variables (we also do this to supply passwords without putting them in hardwired config).

While it's possible to provide a Redis URL using the `{:system, url}` magic, it's kind of clunky trying to munge the individual bits of config (SSL, password, server etc) into another environment variable before startup, and seems like pointless effort when internally Redix is just going to split them all up again. Plus it still doesn't allow any other bits (TTL etc) to be configured dynamically.

There's generally two ways to fix this that I've seen - my preferred one is to add generalised support using [Confex](https://github.com/Nebo15/confex). That requires no changes to the config spec, but allows each individual field to be specified as `{:system, :type, "ENV_VAR", default}`.

However some people don't like having an added dependency in their library, so an alternative is to allow something in the config like `{:callback, Module, :function}` and have the application provide a function which returns config.

I'm happy to implement whichever of those you prefer (or some alternative) - just let me know.

tompave:
Hi, thank you for the interest in the library.

I'm aware of the limitation, but before addressing it I was waiting for Elixir to take a stance on runtime configuration, which has been planned for a while. It has now landed in the recently released 1.9.

More in general, I think at this point I'd like to try to add support for the `init` callbacks that the Elixir core team has been recommending. That would be a customization entry point for all the configuration of the library, which seems handy.

In the meantime, to get you unlocked, I suppose that either `Confex` or [`deferred_config`](https://hex.pm/packages/deferred_config) could solve the problem in your application.

PR #48, opened 2019-06-27
Add Confex configuration support

bernardd:
As discussed over in #47, this change adds Confex support to allow for more flexible runtime configuration.

### DIFF
--- a/lib/fun_with_flags/config.ex
+++ b/lib/fun_with_flags/config.ex
@@ -22,7 +22,7 @@ defmodule FunWithFlags.Config do
   ]
 
   def redis_config do
-    case Application.get_env(:fun_with_flags, :redis, []) do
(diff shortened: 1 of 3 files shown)


tompave:
Hi, thank you for the PR, but I don't think that I can accept it, sorry.

In https://github.com/tompave/fun_with_flags/issues/47 I said that you could try to use either `confex` or `deferred_config` _in your application_ to solve the problem, but I didn't mean that I'm planning to add another dependency to this library. Apologies if that was not clear.

I'm not super familiar with `confex`, so I'm not sure how easy it would be to use it in your application, but `deferred_config` should just work. It will simply modify the application env at boot-time.

bernardd:
No worries. I'll just keep it on a fork until we move to Elixir's release system then I can use their dynamic config - thanks for the library :)

PR #49, opened 2019-07-02
Remove redundant parts of specs

LostKobrakai:
`[type]` includes empty list as value, only `[...]` and `[type, ...]` denote non empty lists.


tompave:
Thanks 👍

PR #50, opened 2019-07-09
Adds InMemory adapter for testing purposes

naps62:
We're currently using this adapter in our own codebase, to avoid extra dependencies within unit tests

It's probably not ready yet, as I've mostly moved the adapter to this repo. I was having some troubles setting up a local test suite (for some reason I can't connect to my local redis instance), so I'm just opening this preemptively.

I'd like to first know if it's an addition worth considering before taking the time to fix the build, if needed :slightly_smiling_face:

### DIFF
--- a/lib/fun_with_flags/store/persistent/in_memory.ex
+++ b/lib/fun_with_flags/store/persistent/in_memory.ex
@@ -1,0 +1,163 @@
+defmodule FunWithFlags.Store.Persistent.InMemory do
+  alias FunWithFlags.{Config, Gate, Flag, Store.Persistent.Ecto.Record}
+
+  def worker_spec do
+    import Supervisor.Spec, only: [worker: 3]
+
+    worker(__MODULE__, [], restart: :permanent)
+  end
+
+  def child_spec(opts \\ []) do
+    %{
+      id: __MODULE__,
+      start: {__MODULE__, :start_link, [opts]}
+    }
+  end
+
+  def start_link(opts \\ []) do
+    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+  end
+
+  def init(_) do
+    {:ok, []}
+  end
+
+  def get(flag_name) do
+    GenServer.call(__MODULE__, {:get, flag_name})
+  end
+
+  def put(flag_name, gate) do
+    GenServer.call(__MODULE__, {:put, flag_name, gate})
+  end
+
+  def delete(flag_name, %Gate{} = gate) do
+    GenServer.call(__MODULE__, {:delete, flag_name, gate})
+  end
+
+  def delete(flag_name) do
+    GenServer.call(__MODULE__, {:delete, flag_name})
+  end
+
+  def all_flags do
+    GenServer.call(__MODULE__, :all_flags)
+  end
+
+  def all_flag_names do
+    GenServer.call(__MODULE__, :all_flag_names)
+  end
+
+  def handle_call({:get, flag_name}, _from, gates) do
+    {:reply, do_get(flag_name, gates), gates}
+  end
+
+  def handle_call({:put, flag_name, gate = %Gate{type: type}}, _from, gates) do
+    record = %Record{
+      flag_name: to_string(flag_name),
+      gate_type: to_string(type),
+      target: Record.serialize_target(gate.for),
+      enabled: gate.enabled
+    }
+
+    index =
+      Enum.find_index(
+        gates,
+        &(&1.flag_name == to_string(flag_name) and
+            &1.gate_type == "percentage")
+      )
+
+    new_gates =
+      case index do
+        nil ->
+          [record | gates] |> Enum.uniq()
+
+        index ->
+          List.replace_at(gates, index, record)
+      end
+
+    publish_change(flag_name)
+
+    {:reply, do_get(flag_name, new_gates), new_gates}
+  end
+
+  def handle_call({:delete, flag_name, gate}, _from, gates) do
+    new_gates =
+      gates
+      |> Enum.reject(
+        &(&1.flag_name == to_string(flag_name) and
+            &1.gate_type == to_string(gate.type) and
+            &1.target == Record.serialize_target(gate.for))
+      )
+
+    publish_change(flag_name)
+
+    {:reply, do_get(flag_name, new_gates), new_gates}
+  end
+
+  def handle_call({:delete, flag_name}, _from, gates) do
+    new_gates =
+      gates
+      |> Enum.reject(&(&1.flag_name == to_string(flag_name)))
+
+    publish_change(flag_name)
+
+    {:reply, do_get(flag_name, new_gates), new_gates}
+  end
+
+  def handle_call(:all_flags, _from, gates) do
+    flags =
+      gates
+      |> Enum.group_by(& &1.flag_name)
+      |> Enum.map(fn {name, records} -> Flag.new(to_atom(name), records) end)
+
+    {:reply, {:ok, flags}, gates}
+  end
+
+  def handle_call(:all_flag_names, _from, gates) do
+    names =
+      gates
+      |> Enum.map(&String.to_atom(&1.flag_name))
+      |> Enum.uniq()
+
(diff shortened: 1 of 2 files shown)


tompave:
Hello, thank you for using the library and for the pull request!

I understand why this adapter can be useful, but I'm not sure if I can add it to the package, sorry.

The reason is that I wouldn't use an in-memory "persistent" store in the test environment, because it wouldn't give me the confidence that everything is working properly. I subscribe to the school of thought that if your application uses a datastore at runtime, then it's better to exercise that dependency in the tests as much as it's reasonable. For example, if in production FWF shares a DB with the application, you'd want it to be shared also in dev and test.

Another consideration that comes to mind is resetting the data between tests. If your test suite uses this in-memory adapter, I'd recommend that it would also expose an API to clear its data. That's something that you'd get for free if you used the Ecto adapter (as it wraps every test in a DB transaction, automatically rolled back), and that it would be easy to implement with the Redis adapter (example steps: [one](https://github.com/tompave/fun_with_flags/blob/v1.4.0/test/support/test_utils.ex#L28), [two](https://github.com/tompave/fun_with_flags/blob/v1.4.0/test/fun_with_flags/store_test.exs#L15) and [three](https://hexdocs.pm/ex_unit/1.9.0/ExUnit.Callbacks.html#on_exit/2)).

I appreciate that what I wrote could be more relevant to integration and acceptance tests rather than unit tests, especially if you're trying to make your unit tests as fast as possible. What I'd do, in that case, is to mock the entire FWF. So, if for example you have one module/function whose logic depends on a FWF flag, then you'd inject FWF and replace it with a mocked implementation in your unit tests for that module and function.

For the reasons above, I feel that adding a builtin in-memory adapter would encourage users to test their applications in a way that I find less than optimal.

However, if you want to proceed with this, I suppose you could release it as an independent extension to FunWithFlags on Hex. All you'd need is to provide the module.

If you want to do that, I have some suggestions on the current implementation. I'll add some comments to the diff. 👍

lib/fun_with_flags/store/persistent/in_memory.ex
```diff
@@ -0,0 +1,163 @@
+defmodule FunWithFlags.Store.Persistent.InMemory do
+  alias FunWithFlags.{Config, Gate, Flag, Store.Persistent.Ecto.Record}
+
```
tompave: It would be good to reference the Behaviour that was [released in v1.4.0](https://github.com/tompave/fun_with_flags/blob/master/CHANGELOG.md#v140):

```elixir
@behaviour FunWithFlags.Store.Persistent
```

lib/fun_with_flags/store/persistent/in_memory.ex
```diff
@@ -0,0 +1,163 @@
+defmodule FunWithFlags.Store.Persistent.InMemory do
+  alias FunWithFlags.{Config, Gate, Flag, Store.Persistent.Ecto.Record}
```
tompave: If this is an in-memory adapter, then it shouldn't use a module scoped under `...Ecto`. That could even not be available in theory, [because](https://github.com/tompave/fun_with_flags/blob/v1.4.0/lib/fun_with_flags/store/persistent/ecto/record.ex#L1-L3):

```elixir
if Code.ensure_loaded?(Ecto) do
  defmodule FunWithFlags.Store.Persistent.Ecto.Record do
```

lib/fun_with_flags/store/persistent/in_memory.ex
```diff
@@ -0,0 +1,163 @@
...
+
+    worker(__MODULE__, [], restart: :permanent)
+  end
+
+  def child_spec(opts \\ []) do
+    %{
+      id: __MODULE__,
+      start: {__MODULE__, :start_link, [opts]}
+    }
+  end
```
tompave: I'm wondering why you need both `worker_spec` and `child_spec`. Some comments would be helpful!

lib/fun_with_flags/store/persistent/in_memory.ex
```diff
@@ -0,0 +1,163 @@
...
+  def all_flag_names do
+    GenServer.call(__MODULE__, :all_flag_names)
+  end
+
+  def handle_call({:get, flag_name}, _from, gates) do
+    {:reply, do_get(flag_name, gates), gates}
+  end
+
+  def handle_call({:put, flag_name, gate = %Gate{type: type}}, _from, gates) do
+    record = %Record{
```
tompave: I don't think you need this. The [behaviour](https://github.com/tompave/fun_with_flags/blob/v1.4.0/lib/fun_with_flags/store/persistent.ex#L38-L39) says that these function only need to return `FunWithFlags.Flag{}` structs. You can store them directly in the GenServer with no need to transform them back and forth into `Record` structs.

lib/fun_with_flags/store/persistent/in_memory.ex
```diff
@@ -0,0 +1,163 @@
...
+
+  defp do_deserialize_gate(%Record{
+          gate_type: "group",
+          enabled: enabled,
+          target: target
+        }) do
+    %Gate{type: :group, for: target, enabled: enabled}
+  end
+
+  defp publish_change(flag_name) do
```
tompave: Starting with [v1.4.0](https://github.com/tompave/fun_with_flags/blob/master/CHANGELOG.md#v140), persistence adapters are not responsible for publishing changes anymore. This function and its calls can be removed.

naps62:
@tompave thanks for the comments.
I fully understand your reasons to not want this. I actually subscribe to the same school of thought, about using the same data stores in tests as much as possible.

To give a bit more context: the reason I implemented this in my own tests was precisely to speed up unit tests, and allow us to keep `async: true` within them.
Mocking FWF entirely would pretty much achieve the same effect, but we would end up with a similar GenServer to keep state for each test

The reset mechanism you mentioned didn't exist because for each test, I'm concerned only with one or two flags at most, and we just do `on_exit fn -> FunWithFlags.clear(flag) end`.

Like I said, this is being useful for us. I'll probably create a separate package as you said, but only because we're starting to use this in multiple projects, and I'd prefer to avoid copy & paste.

Thanks :ok_hand:

lib/fun_with_flags/store/persistent/in_memory.ex
naps62: yup, I was actually investigating this around the same time I opened this PR.
We had our own fork of FWF precisely to remove those checks, because we were having occasional compilation issues.

On further inspection, I think it's because I was sometimes doing `mix deps.compile --force fun_with_flags`. But I didn't know about the `--include-children`. So FWF would end up not being compiled with ecto support.

lib/fun_with_flags/store/persistent/in_memory.ex
naps62: good to know :ok_hand:

lib/fun_with_flags/store/persistent/in_memory.ex
naps62: the worker_spec is probably a left over from when I was first doing this (copying from the Ecto Adapter), which I forgot to remove

lib/fun_with_flags/store/persistent/in_memory.ex
naps62: if I recall correctly, this was done because I needed to use the Ecto Serializer, which is only implemented for `Ecto.Record`. Not doing so would cause inconsistencies between tests and production (atoms vs strings, etc)

lib/fun_with_flags/store/persistent/in_memory.ex
naps62: good to know as well, thanks :ok_hand:

tompave:
> We had our own fork of FWF precisely to remove those checks
(...)
> Like I said, this is being useful for us. I'll probably create a separate package as you said, but only because we're starting to use this in multiple projects, and I'd prefer to avoid copy & paste.

Ok @naps62, sounds like a plan. My only recommendation is to not release the package publicly if it can only work with your fork, because it wouldn't be useful for anyone else. Ideally it should work with the latest released version of FWF.

PR #54, opened 2019-11-13
Improve error handling for redis store

chubarovNick:


tompave:
Hello!

Thank you for the PR!

Can you please provide some more info on why you needed these changes? An example of the cases where these errors would happen?

tompave:
I get `%Redix.ConnectionError{}`, but I'd be curious about the `%Redix.Error{}`. If the Redis adapter is sending invalid command to the Redis server, then that's something that should be fixed.

tompave:
Is it something like:

```
set fun_with_flags:foobar just_a_string
```

and then:

```elixir
FunWithFlags.enabled?(:foobar)
** (Protocol.UndefinedError) protocol String.Chars not implemented for %Redix.Error{message: "WRONGTYPE Operation against a key holding the wrong kind of value"} of type Redix.Error (a struct).
```

chubarovNick:
Hello. We face `%Redix.ConnectionError{}` in our project, and then I handle all possible error values from `Redix.command/2`. I think it's almost impossible to get this error with your library :)

tompave:
I see, fair enough.

Thank you for the improvement!

Issue #56, opened 2019-12-04
Spec definition and documentation mismatch


skylerparr:
Thank you for creating this library. I use it often. I recently added rules for groups and the dialyzer caught that I wasn't using atoms for groups. On further investigation it looks as though FunWithFlagsUI inserts the group rule into the database as a String and is passed to the `in?` function as a String (https://github.com/tompave/fun_with_flags/blob/master/lib/fun_with_flags/protocols/group.ex#L68). But, the specs specify that the group should be an `atom`. As far as I can tell it comes from the database unchanged: 
```
iex(99)> FunWithFlags.get_flag(:bar)
2019-12-04 14:11:39.911 [info] [application=ecto_sql module=Ecto.Adapters.SQL function=log/4 pid=<0.1619.0> ] QUERY OK source="fun_with_flags_toggles" db=0.7ms
SELECT DISTINCT f0.`flag_name` FROM `fun_with_flags_toggles` AS f0 []
%FunWithFlags.Flag{
  gates: [
    %FunWithFlags.Gate{enabled: false, for: nil, type: :boolean},
    %FunWithFlags.Gate{
      enabled: true,
      for: "inserted_at<2019-11-8",
      type: :group
    }
  ],
  name: :bar
}
```
The documentation also shows the usage as strings (https://github.com/tompave/fun_with_flags#group-gate). My guess is that the spec should be `atom() | String.t()` or `term()`.

tompave:
Hello!

Thank you for using the library and for opening the issue!

Group names used to be atoms, but a couple of years ago they were changed to binaries in `v0.10.0`: https://github.com/tompave/fun_with_flags/blob/master/CHANGELOG.md#v0100

You're right, the spec is wrong. I assume that this is the offending line: https://github.com/tompave/fun_with_flags/blob/v1.4.1/lib/fun_with_flags/protocols/group.ex#L68

It should indeed be `String.t() | atom()`. Also, the `@doc """` above the protocol method definition uses atoms, but it should be binaries.

Would you be happy to send a PR? 🙂

tompave:
Solved with the PR.

PR #57, opened 2019-12-05
Fixed spec on group protocol

skylerparr:
ref: https://github.com/tompave/fun_with_flags/issues/56

### DIFF
--- a/lib/fun_with_flags/protocols/group.ex
+++ b/lib/fun_with_flags/protocols/group.ex
@@ -65,7 +65,7 @@ defprotocol FunWithFlags.Group do
       iex> FunWithFlags.Group.in?(data, :superusers)
       false
   """
-  @spec in?(term, atom) :: boolean
+  @spec in?(term, atom | String.t) :: boolean
   def in?(item, group)
 end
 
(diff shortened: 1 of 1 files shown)


lib/fun_with_flags/protocols/group.ex
```diff
@@ -65,7 +65,7 @@ defprotocol FunWithFlags.Group do
       iex> FunWithFlags.Group.in?(data, :superusers)
       false
   """
-  @spec in?(term, atom) :: boolean
+  @spec in?(term, atom | String.t) :: boolean
```
tompave: Can we make this `String.t | atom` please? Mainly because this is really meant to be a binary, and atoms are supported for retro-compatibility.

Also, the documentation comment above this function still uses atoms (`:staff`, `:superusers`). It would be great if you could make them binaries too.

skylerparr: Sorry for the delay. Got it updated. What do you think?

tompave: Perfect, thanks!

PR #60, opened 2019-12-17
Deterministic TTL flutter proof of concept

kieraneglin:
This is in response to the TODO item of "Add some optional randomness to the TTL".  I'm open to different naming suggestions.

This PR adds a deterministic negative offset to the TTL for a flag.  This offset is computed based off the flag's name, is always < 10% of the TTL, and is always negative so the TTL becomes the _maximum_ life of a flag in cache.

This may seem like a strange approach, but I ran into some trouble with introduction of a purely random offset ([example here](https://github.com/kieraneglin/fun_with_flags/blob/ttl-flutter/lib/fun_with_flags/config.ex#L42)).  With an approach like this, rapid attempts to read the cache (and therefor rapid checking of expiration while considering a random offset) would tend to expire the cache at the lower bound of the offset, defeating the purpose.  This also gets messy when the TTL is saved as a module attribute (`@ttl Config.cache_ttl`), since that gets locked in at compile time and removes any randomness.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ erl_crash.dump
 
 # Redis db
 dump.rdb
+
+# The directory used by Elixir Language Server
+.elixir_ls
--- a/lib/fun_with_flags/flag.ex
+++ b/lib/fun_with_flags/flag.ex
@@ -6,6 +6,7 @@ defmodule FunWithFlags.Flag do
   """
 
   alias FunWithFlags.Gate
+  alias FunWithFlags.Config
 
   defstruct [name: nil, gates: []]
   @type t :: %FunWithFlags.Flag{name: atom, gates: [FunWithFlags.Gate.t]}
@@ -44,6 +45,31 @@ defmodule FunWithFlags.Flag do
             check_boolean_gate(gates) || check_percentage_gate(gates, item, flag_name)
         end
     end
+  end
+
+  @doc ~S"""
+  Taking a %Flag{}, a deterministic TTL offset is returned that is within 10% of the default TTL.
+
+  This number will be <= 0, meaning that the TTL with flutter will only ever be <= the original TTL.
+  """
+  @spec flutter_offset(t) :: integer
+  def flutter_offset(%__MODULE__{name: flag_name}) do
+    flutter_percentage = 0.1
+    maximum_ttl_variance = ceil(Config.cache_ttl * flutter_percentage)
+
+    flag_name
+    |> name_as_integer()
+    |> Integer.mod(maximum_ttl_variance)
+    |> Kernel.*(-1)
+  end
+
+  defp name_as_integer(flag_name) do
+    {name_as_integer, _} =
+      :crypto.hash(:md5, Atom.to_string(flag_name))
+      |> Base.encode16()
+      |> Integer.parse(16)
+
+    name_as_integer
   end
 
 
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -21,3 +21,6 @@ erl_crash.dump
 
 # Redis db
 dump.rdb
+
+# The directory used by Elixir Language Server
+.elixir_ls
```
kieraneglin: Can remove if needed

lib/fun_with_flags/flag.ex
```diff
@@ -46,6 +47,31 @@ defmodule FunWithFlags.Flag do
...
+
+    flag_name
+    |> name_as_integer()
+    |> Integer.mod(maximum_ttl_variance)
+    |> Kernel.*(-1)
+  end
+
+  defp name_as_integer(flag_name) do
+    {name_as_integer, _} =
+      :crypto.hash(:md5, Atom.to_string(flag_name))
```
kieraneglin: This approach was taken to introduce some level of randomness, since flags that were similar in name would tend to be clumped together in their offset

.gitignore
tompave: Let's keep it, it seems sensible to have it here.

tompave:
Hi @kieraneglin, I've merged into master some changes to how the cached data is stored in the ETS table. This should make it a bit easier to adapt your implementation in the way I suggested.

kieraneglin:
@tompave Thank you!  I'll review and modify as-needed.  And thank you for the thoughtful code review!  I hope to integrate these changes shortly after the holidays

lib/fun_with_flags/store/cache.ex
```diff
@@ -3,12 +3,12 @@ defmodule FunWithFlags.Store.Cache do
   use GenServer
   alias FunWithFlags.Timestamps
   alias FunWithFlags.Flag
+  alias FunWithFlags.Config
 
   @table_name :fun_with_flags_cache
   @table_options [
     :set, :protected, :named_table, {:read_concurrency, true}
   ]
-  @ttl FunWithFlags.Config.cache_ttl
```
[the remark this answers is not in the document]
tompave: This comment is not relevant anymore. The cache TTL is not stored in the ETS table alongside the cached data, since https://github.com/tompave/fun_with_flags/commit/effe7a74c964d61f9f3993caae39ccb80f04c4a8.

lib/fun_with_flags/store/cache.ex
```diff
@@ -41,7 +41,7 @@ defmodule FunWithFlags.Store.Cache do
   end
 
   defp validate(name, flag = %Flag{name: name}, timestamp) do
-    if Timestamps.expired?(timestamp, @ttl) do
+    if flag_stale?(timestamp, flag) do
```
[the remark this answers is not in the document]
tompave: Those changes have now been merged into master with https://github.com/tompave/fun_with_flags/commit/effe7a74c964d61f9f3993caae39ccb80f04c4a8.

This means that the adjusted (fluttered?) TTL can be calculated when storing the data into the cache.

tompave:
Hi @kieraneglin, are you still interested in completing this?

kieraneglin:
@tompave I am still interested!  Apologies on the delays.

If you think there's still value in refactoring this PR, I can pick it back up.

tompave:
Hi @kieraneglin, no problem :-)

Sure, I think it would be valuable. Thanks!

tompave:
Hi @kieraneglin, are you still interested in finishing this?

tompave:
Hey @kieraneglin, I'm closing this because of inactivity. Please re-submit it if you want to pick it up again.

PR #64, opened 2020-03-09
Add ability to configure toggle table name

### LINKED ISSUE #62: Support for custom table names
right now, it seems like the [table name is hardcoded to "fun_with_flags_toggles"](https://github.com/tompave/fun_with_flags/blob/3b5a1c88436fd739858051268f9af298fc0337ae/lib/fun_with_flags/store/persistent/ecto/record.ex). It'd be nice to be able to use a different name

BobbyMcWho:
Resolves #62

I'd love suggestions on how to best represent this in tests, as well as any general feedback. I'm relatively new to Elixir packages and their best practices for things like configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -671,7 +671,7 @@ The Mixfile defines a few other helper tasks that allow to run the test suite wi
 ### Configuration changes have no effect in `MIX_ENV=dev`
 
 **Issue**: changing the library settings in the host application's Mix config file has no effect, or "missing process" exceptions are raised when booting. This should only be an issue in the development environment.  
-**Solution**: clear the compiled BEAM bytecode with:  
+**Solution**: clear the compiled BEAM bytecode with:
 
 ```shell
 rm -r _build/dev/lib/fun_with_flags
--- a/README.md
+++ b/README.md
@@ -551,7 +551,8 @@ config :fun_with_flags, :persistence,
   adapter: FunWithFlags.Store.Persistent.Ecto,
   repo: MyApp.Repo
 
-# Optional configuration to specify the database table name. Default table name is "fun_with_flags_toggles".
+# Optional configuration to specify the ecto table name.
+# Default table name is "fun_with_flags_toggles".
 config :fun_with_flags, table_name: "your_table_name"
 ```
 
--- a/README.md
+++ b/README.md
@@ -503,7 +503,7 @@ config :fun_with_flags, :cache_bust_notifications,
   [enabled: true, adapter: FunWithFlags.Notifications.Redis]
 
 # Notifications can also be disabled, which will also remove the Redis/Redix dependency
-config :fun_with_flags, :cache_bust_notifications, [enabled: false]
+config :fun_with_flags, :cache_bust_notifications, [enabled: false]  
 ```
 
 When using Redis for persistence and/or cache-busting PubSub it is necessary to configure the connection to the Redis instance. These options can be omitted if Redis is not being used. For example, the defaults:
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -560,7 +563,7 @@ The library comes with two PubSub adapters for the [`Redix`](https://hex.pm/pack
 
 The Redis PubSub adapter is the default and doesn't need to be excplicity configured. It can only be used in conjunction with the Redis persistence adapter however, and is not available when using Ecto for persistence. When used, it connects directly to the Redis instance used for persisting the flag data.
 
-The Phoenix PubSub adapter uses the high level API of `Phoenix.PubSub`, which means that under the hood it could use either its PG2 or Redis adapters, and this library doesn't need to know. It's provided as a convenient way to leverage distributed Erlang when using FunWithFlags in a Phoenix application, although it can be used independently (without the rest of the Phoenix framework) to add PubSub to Elixir apps running on Erlang clusters.
```
tompave: Can we restore these trailing spaces please? Trailing spaces in markdown are meaningful. Also in the code snipped above please, because this PR does not relate to that example.

lib/fun_with_flags/config.ex
```diff
@@ -32,6 +32,10 @@ defmodule FunWithFlags.Config do
     end
   end
 
+  def table_name do
+    Application.get_env(:fun_with_flags, :table_name, "fun_with_flags_toggles")
```
tompave: I would put `"fun_with_flags_toggles"` in a module attribute:

```elixir
@default_ecto_table_name "fun_with_flags_toggles"
```

README.md
BobbyMcWho: Sure, I will have to figure out how to make my ide let me do that 😆

lib/fun_with_flags/config.ex
tompave: Also, the concept of `table_name` should really be namespaced under the `:persistence` group, and clearly reference ecto. So this could be:

```elixir
@default_persistence_config [
  adapter: FunWithFlags.Store.Persistent.Redis,
  repo: FunWithFlags.NullEctoRepo,
  ecto_table_name: @default_ecto_table_name
]

# ...
def ecto_table_name do
  Keyword.get(persistence_config(), :ecto_table_name)
end
```

And then be configured with:

```elixir
config :fun_with_flags, :persistence,
  adapter: FunWithFlags.Store.Persistent.Ecto,
  repo: FunWithFlags.Dev.EctoRepo,
  ecto_table_name: "my_table"
```

tompave: I know that `repo` is not namespaced as `ecto_repo`, but:

1. Changing that now would be a big breaking change.
2. In a Phoenix and/or Ecto app, "repo" is almost universally understood as an Ecto repo. On the other hand "table" could be an ETS table, or some people might think it affects Redis somehow. I'd rather avoid ambiguities.

priv/ecto_repo/migrations/00000000000000_create_feature_flags_table.exs
```diff
@@ -1,8 +1,11 @@
 defmodule FunWithFlags.Dev.EctoRepo.Migrations.CreateFeatureFlagsTable do
   use Ecto.Migration
+  alias FunWithFlags.Config
+
+  @table_name Config.table_name() |> String.to_atom()
```
tompave: I wouldn't make this dynamic in the migrations. The reason is that migrations should be as simple, logic-free, and dependency-free as possible. They should reflect the state of the DB at the moment in time the migration was created and run. Making it dynamic, here, would mean that the migration would have a different effect later, if run again after the config has changed.

I think adding a comment would be enough. Something that says that the table name can be optionally customized in the config, and that in that case the migration should reflect that.

priv/ecto_repo/migrations/00000000000001_ensure_columns_are_not_null.exs
```diff
@@ -10,8 +11,10 @@ defmodule FunWithFlags.Dev.EctoRepo.Migrations.EnsureColumnsAreNotNull do
   # is no need to run this migration. In that case, this migration
   # is a no-op.
 
+  @table_name Config.table_name()
+
   def up do
-    alter table(:fun_with_flags_toggles) do
+    alter table(@table_name) do
```
tompave: Here too. No need for this change.

README.md
```diff
@@ -550,6 +550,9 @@ config :my_app, MyApp.Repo,
 config :fun_with_flags, :persistence,
   adapter: FunWithFlags.Store.Persistent.Ecto,
   repo: MyApp.Repo
+
+# Optional configuration to specify the database table name. Default table name is "fun_with_flags_toggles".
+config :fun_with_flags, table_name: "your_table_name"
```
BobbyMcWho: @tompave you had mentioned this code block, but I think it makes sense in context? 
image

BobbyMcWho:
@tompave I believe I have addressed all of your feedback, I appreciate your time reviewing this :)

tompave:
Awesome @BobbyMcWho , great work!
I just need to figure out why the CI checks are not being reported for this PR. (they normally work)

Travis is building the PR [here](https://travis-ci.org/tompave/fun_with_flags/builds/660815340), but it looks odd because it seems to reference a commit that doesn't exist in the PR: https://github.com/tompave/fun_with_flags/commit/01ef390c41eab84d6ff7409eb68000aa8a0c2f60

In the meantime, can you please add a test for the new config getter? Something like this: https://github.com/tompave/fun_with_flags/blob/v1.4.1/test/fun_with_flags/config_test.exs#L42-L53 (and with a matching `reset_persistence_defaults/0` function). Thanks!

BobbyMcWho:
It looks like there are 80 permutations of erlang/elixir/env. This makes the tests take a while. Is there any way to re-use some of these containers so that we don't have to spin them up again?

BobbyMcWho:
@tompave It seems like CI is failing because it can't pull that version of erlang on bionic for whatever reason

tompave:
Hi @BobbyMcWho sorry for losing track of this for a while.

I've update the TravisCI config for the repo, plus updated some deps on master. Can you please merge in master and push?

tompave:
I'm merging this into another feature branch in this main repo, so that I can try to unblock the CI checks.

PR #65, opened 2020-03-11
Cache deps and _build

BobbyMcWho:
Not sure if this is still best practice or recommended: https://dockyard.com/blog/2015/12/02/speeding-up-elixir-project-build-times-on-travis-ci

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,3 +38,7 @@ notifications:
   email:
     on_success: never
     on_failure: never
+cache:
+  directories:
+    - _build
+    - deps


tompave:
Hi, sorry for the long wait and for sending this improvement.

You're right, the build was painfully slow.
I've already removed an accidental double-compilation issue with this commit: https://github.com/tompave/fun_with_flags/commit/0c3ec15a26e540df720c3e893e36e624c42ed693. That reduced each job's time by almost half! (was 2m10s on average, now 1m10s)

With this it should get even faster.

PR #71, opened 2020-04-20
Simplify the travis setup so that it runs more tests per job

tompave:


tompave:
Closing, because this setup ignores some failures.

PR #92, opened 2021-01-30
enabled? for multiple flags

elpikel:
Added possibility to check if more than one flag is enabled at the same time.

Most of changes are because of mix format.


tompave:
Hey there, thank you for submitting the PR.

A couple of comments.

1. This project is not mix-formatted, and PRs that format the entire codebase with `mix format` won't be accepted. The reason is that it's time consuming to review such diffs, and for security reasons I'm not comfortable merging contributions that I can't review throughly (what if something bad is hidden somewhere in the mix-format diff?). Maybe one day I'll mix-format the project myself.
2. Aside from my comment on mix-format in isolation, the way it's done in this PR makes reviewing the changes difficult. There is a "mix format" commit (https://github.com/tompave/fun_with_flags/pull/92/commits/0f0e72eb97dea532fce3cfafefe4c2f6462c8f20), preceded by a first commit (https://github.com/tompave/fun_with_flags/pull/92/commits/8c6fd2fcf41b338711a9ae9d4518972673323342) that seems to introduce the actual logic changes. However, that first commit also seems to include the mix-formatting of several files. That makes it difficult for me to isolate and review the actual changes introduced in that commit.

With that in mind, I would request that any mix-format changes are removed from the PR. You can either close this and re-open it, or change things in the branch history and force-push.

That being said, keep in mind that this is a significant change in the public interface of the package, so there is a chance that I won't accept it anyway. A design principle of this package is simplicity, and I would have to be convinced that checking multiple flags at once provides advantages over checking them together with a boolean `AND`.

PR #93, opened 2021-01-31
extended enabled? to accept multiple flags

elpikel:
Added possibility to check if more than one flag is enabled at the same time.

### DIFF
--- a/lib/fun_with_flags.ex
+++ b/lib/fun_with_flags.ex
@@ -37,6 +37,7 @@ defmodule FunWithFlags do
 
   It can be invoked with just the flag name, as an atom,
   to check the general status of a flag (i.e. the boolean gate).
+  It can also be invoked with list of flags names where each should be an atom.
 
   ## Options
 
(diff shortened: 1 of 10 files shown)


tompave (commented):
Hi, thank you for submitting this PR and for cleaning it after my initial feedback (https://github.com/tompave/fun_with_flags/pull/92#issuecomment-770378188).

I can see how this would work, but I'm not convinced that adding it to the package would be right, sorry. I feel that it adds a non-trivial amount of complexity that outweighs any benefit of adding that new flag-query style.

As for the benefits, I can think of UX conveniency and query performance.

With regard to the UX, I'm not convinced that this change is really necessary, because multi-flag checks can be achieved easily with the current API:

```elixir
FunWithFlags.enabled?(:foo) && FunWithFlags.enabled?(:bar)
```

It's actually even better. Because of the simplicity of the current interface, it is easily composable and it provides the flexibility to test different actors and groups for different flags:

```elixir
import FunWithFlags
def is_feature_enabled?(user, country) do
  enabled?(:new_feature, for: country) || enabled?(:new_feature, for: user) || enabled?(:admin_flag_overrides, for: user)
end

if is_feature_enabled?(current_user, country_for_session(conn)) do
  # ...
end
```

With regard to the other benefit, query performance, I can see that this PR would indeed make this better for Ecto, as multiple flags would be loaded together instead of performing several queries in sequence.
However, the package would only take advantage of this improvement in the case of cache misses. With a warm cache, everything would be read from cache anyway. My gut feeling is that the real-world performance benefits of this group query would not be apparent very often.
Yes, I can conceive scenarios where a large number of flags is checked at the same time, maybe with the cache disabled, but I think that that's a fairly uncommon case, and it doesn't justify the added complexity of this change.
If a user of the package really needs to avoid the sequence of queries, and they don't use a cache anyway, they can try with `FunWithFlags.all_flags/0` and then check them manually.

---

Even though I'm not accepting the PR, I have left some comments to suggest a different approach for the entry point to this functionality. I've done so because I appreciate the time that was spent on this, and I would like to articulate my thoughts on simplicity as a design goal. I think that separate functions for the "many flags" functionality would be better than overloading and extending the current ones (both for the public interface and the internal modules). By adding more multi-flag clauses with recursion to the single-flag clauses the complexity of the functions increases considerably, whereas a new set of functions would help keeping things simple.

PR #102, opened 2021-03-03
Notifications into the supervisor

### LINKED ISSUE #31: Notifications need to start with the application supervisor
The Notification module has to wait on the Application that contains the PubSub to start otherwise it fails, now this is handled with retries but they break really easily and littler tests.

I know that the notifications are now optional but I think the right approach is just to have the Application supervisor that spawns the PubSub also spawn the the Notification module.

Here are the errors: 

```
FunWithFlags: Cannot subscribe to Phoenix.PubSub process MyApp.PubSub (exception: %ArgumentError{message: "argument error"})
FunWithFlags: retrying to subscribe to Phoenix.PubSub, attempt 2
```

mindreader:
Would fix #31 

This puts the phoenix pubsub into the fun with flags supervisor.  That means both of the pubsub related applications start up at roughly the same time, regardless of what your own application does.

It also means you don't have to configure the phoenix pubsub module into your supervision tree, or even specify it in the fun with flags cache busting config (it defaults to an obvious choice).  You can supply options to phoenix pub sub pretty easily, and for people who don't want to do this, the old option still works.

This would break people who have put the pubsub process into their own supervision tree, not sure what to do about that.  I also am not using phoenix, so not totally sure that works but it probably does.  This is probably pretty close to a solution if you have any feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -585,7 +585,7 @@ config :fun_with_flags, :cache_bust_notifications,
 Or, without Phoenix:
 
 ```elixir
-# possibly in the application's supervision tree
+# possibly in the application's supervision tree, alternatively specified below
 children = [
   {Phoenix.PubSub, [name: :my_pubsub_process_name, adapter: Phoenix.PubSub.PG2]}
 ]
@@ -597,6 +597,9 @@ config :fun_with_flags, :cache_bust_notifications,
   enabled: true,
   adapter: FunWithFlags.Notifications.PhoenixPubSub,
   client: :my_pubsub_process_name
+  # pubsub_driver: Phoenix.PubSub.PG2 (optional)
+  # pubsub_driver: {Phoenix.PubSub.PG2, pool_size: 10} (optional with options)
+
 ```
 
 ## Installation
--- a/lib/fun_with_flags/application.ex
+++ b/lib/fun_with_flags/application.ex
@@ -10,12 +10,11 @@ defmodule FunWithFlags.Application do
     Supervisor.start_link(children(), opts)
   end
 
-
   defp children do
     [
       FunWithFlags.Store.Cache.worker_spec(),
-      persistence_spec(),
-      notifications_spec(),
+      persistence_spec() |
+      notifications_spec()
     ]
     |> Enum.reject(&(!&1))
   end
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -597,6 +597,9 @@ config :fun_with_flags, :cache_bust_notifications,
   enabled: true,
   adapter: FunWithFlags.Notifications.PhoenixPubSub,
   client: :my_pubsub_process_name
+  # pubsub_driver: Phoenix.PubSub.PG2 (optional)
+  # pubsub_driver: {Phoenix.PubSub.PG2, pool_size: 10} (optional with options)
```
tompave, line 601: I think it would be clearer if the commented-out code was valid Elixir. That mean that the text in parentheses should be its own comment.
```suggestion
  # pubsub_driver: Phoenix.PubSub.PG2                  # optional
  # pubsub_driver: {Phoenix.PubSub.PG2, pool_size: 10} # optional with options
```

lib/fun_with_flags/application.ex
```diff
@@ -42,7 +41,21 @@ defmodule FunWithFlags.Application do
   #
   defp notifications_spec do
     try do
-      Config.change_notifications_enabled? && Config.notifications_adapter.worker_spec()
+      if Config.change_notifications_enabled? do
+
+        {driver, opts} = Config.notifications_pubsub_driver() |> case do
+          mod when is_atom(mod) -> {mod, []}
+          {mod, opts} = x when is_atom(mod) and is_list(opts) -> x
+        end
```
tompave, line 49: I find this a bit difficult to read, how about:
```suggestion
        {driver, opts} =
          case Config.notifications_pubsub_driver() do
            mod when is_atom(mod) ->
              {mod, []}
            {mod, opts} when is_atom(mod) and is_list(opts) ->
              {mod, opts}
            _ ->
              raise "unexpected config for ...."
          end
```

tompave (commented):
Hi, thank you for working on this problem.

If I understand these changes, you're trying to always start a `Phoenix.PubSub.PG2` process, with the name configured in mix config. Basically, you're trying to put the host Phoenix app's pubsub process under the supervision tree of FunWithFlags.

I have a few thoughts:

1. Wouldn't then the host application's supervision tree complain because there are two `Phoenix.PubSub.PG2` processes with the same name?
2. The code in the PR defaults to starting a `Phoenix.PubSub.PG2` process. If the host application is using `Phoenix.PubSub.Redis` instead, would the idea be that they need to configure it as the `:pubsub_driver`? I guess it would work, but I am not convinced by the interface. That's basically two config fields for almost the same thing. I would prefer if it could be inferred.
3. This will always try to start _something_ as the `Config.notifications_pubsub_driver()`. But what if FunWithFlags is configured to use Redis pubsub? In that case nothing would be required. I suppose that the config could set `pubsub_driver: nil`, but it would be a bit awkward.

All in all, I'm not convinced that FunWithFlags should supervise processes that belong to Phoenix. That not playing nicely with the ecosystem and it feels like asking for trouble. I would find it odd if a dependency of a Phoenix app started messing with the supervision tree of the host application, without any explicit opt in.

I'm adding a few more thoughts in a comment to https://github.com/tompave/fun_with_flags/issues/31.

mindreader:
I don't really have good solutions for many of the issues you mentioned, and I knew that this wasn't probably going to be a perfect solution, but I hoped it would be close.  I think the options you gave in the original issue are better options than this, so I will close this until I come up with something better.

PR #103, opened 2021-03-11
Improve control over the start behaviour

tompave:
Addresses https://github.com/tompave/fun_with_flags/issues/31

This PR introduces extracts from the application module all the supervision logic into new `FunWithFlags.Supervisor` module. This new module can be used directly if users of the package want to manage the supervision tree directly. The PR also adds detailed instructions on how to accomplish this.

To inspect the new docs: https://github.com/tompave/fun_with_flags/tree/custom_start#application-start-behaviour

### DIFF
--- a/README.md
+++ b/README.md
@@ -685,14 +685,14 @@ defmodule MyPhoenixApp.Application do
 
 Then it's necessary to configure the Mix project to not start the `:fun_with_flags` application automatically. This can be accomplished in the Mixfile in a number of ways, for example: (**Note**: These are alternative solutions, you don't need to do both. You must decide which is more appropriate for your setup.)
 
-* Declare the `:fun_with_flags` dependency with either the `runtime: false` or `app: false` options. ([docs](https://hexdocs.pm/mix/1.11.3/Mix.Tasks.Deps.html#module-dependency-definition-options))
+* **Option A**: Declare the `:fun_with_flags` dependency with either the `runtime: false` or `app: false` options. ([docs](https://hexdocs.pm/mix/1.11.3/Mix.Tasks.Deps.html#module-dependency-definition-options))
 
 ```diff
 - {:fun_with_flags, "~> 1.6"},
 + {:fun_with_flags, "~> 1.6", runtime: false},
 ```
 
-* Declare that the `:fun_with_flags` application is managed directly by your host application ([docs](https://hexdocs.pm/mix/1.11.3/Mix.Tasks.Compile.App.html)).
+* **Option B**: Declare that the `:fun_with_flags` application is managed directly by your host application ([docs](https://hexdocs.pm/mix/1.11.3/Mix.Tasks.Compile.App.html)).
 
 ```diff
   def application do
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -649,6 +652,62 @@ And then configure the library to use it:
...
+    opts = [strategy: :one_for_one, name: MyPhoenixApp.Supervisor]
+    Supervisor.start_link(children, opts)
+  end
+
+  # ...
+```
+
+Then it's necessary to configure the Mix project to not start the `:fun_with_flags` application automatically. This can be accomplished in the Mixfile in a number of ways, for example: (**Note**: These are alternative solutions, you don't need to do both. You must decide which is more appropriate for your setup.)
+
+* Declare the `:fun_with_flags` dependency with either the `runtime: false` or `app: false` options. ([docs](https://hexdocs.pm/mix/1.11.3/Mix.Tasks.Deps.html#module-dependency-definition-options))
```
scottmessinger, line 688: In reading these, I wasn't sure if these were two steps to one solution or two different solutions. I think this might help clarify things a bit:

```suggestion
* **Option A**: Declare the `:fun_with_flags` dependency with either the `runtime: false` or `app: false` options. ([docs](https://hexdocs.pm/mix/1.11.3/Mix.Tasks.Deps.html#module-dependency-definition-options))
```

README.md
```diff
@@ -649,6 +652,62 @@ And then configure the library to use it:
...
+
+```diff
+- {:fun_with_flags, "~> 1.6"},
++ {:fun_with_flags, "~> 1.6", runtime: false},
+```
+
+* Declare that the `:fun_with_flags` application is managed directly by your host application ([docs](https://hexdocs.pm/mix/1.11.3/Mix.Tasks.Compile.App.html)).
```
scottmessinger, line 695:
```suggestion
* **Option B**: Declare that the `:fun_with_flags` application is managed directly by your host application ([docs](https://hexdocs.pm/mix/1.11.3/Mix.Tasks.Compile.App.html)).
```

README.md
```diff
@@ -649,6 +652,62 @@ And then configure the library to use it:
...
+      mod: {MyPhoenixApp.Application, []},
++     included_applications: [:fun_with_flags],
+      extra_applications: [:logger, :runtime_tools]
+    ]
+  end
+```
+
+The result of those changes is that the `:fun_with_flags` application won't be loaded and started automatically, and therefore the FunWithFlags supervision tree won't risk to be started before the other processes in the host Phoenix application. Rather, the supervision tree will start alongside the other core Phoenix processes.
+
+One final note on this topic is that if you're also using [`FunWithFlags.UI`](https://github.com/tompave/fun_with_flags_ui) (refer to the [Web Dashboard](#web-dashboard) section, above in this document), then that will need to be configured as well. The reason is that `:fun_with_flags` is a dependency of `:fun_with_flags_ui`, so including the latter as a dependency will cause the former to be auto-started despite the configuration described above. To avoid this, the same configuration should be used for the `:fun_with_flags_ui` dependency, regardless of the approach used (`runtime: false`, `app: false`, or `included_applications`).
```
scottmessinger, line 709:
```suggestion
One final note on this topic is that if you're also using [`FunWithFlags.UI`](https://github.com/tompave/fun_with_flags_ui) (refer to the [Web Dashboard](#web-dashboard) section, above in this document), then that will need to be configured as well. The reason is that `:fun_with_flags` is a dependency of `:fun_with_flags_ui`, so including the latter as a dependency will cause the former to be auto-started despite the configuration described above. To avoid this, the same configuration should be used for the `:fun_with_flags_ui` dependency, regardless of the approach used (Option A (`runtime: false`, `app: false`), or Option B (`included_applications`)).
```

scottmessinger (changes requested):
Thanks for doing this! I'll try out the branch, but just some feedback on the readme changes!

README.md
tompave: Thanks, that makes sense. I was hoping that the `**Note**: ...` would help with this, but labelling these as options is better.

README.md
tompave: Makes sense, but I'd rather avoid the nested parentheses.
```suggestion
One final note on this topic is that if you're also using [`FunWithFlags.UI`](https://github.com/tompave/fun_with_flags_ui) (refer to the [Web Dashboard](#web-dashboard) section, above in this document), then that will need to be configured as well. The reason is that `:fun_with_flags` is a dependency of `:fun_with_flags_ui`, so including the latter as a dependency will cause the former to be auto-started despite the configuration described above. To avoid this, the same configuration should be used for the `:fun_with_flags_ui` dependency, regardless of the approach used (Option A: `runtime: false`, `app: false`; or Option B: `included_applications`).
```

scottmessinger: Agreed! I didn't like the double parens either. Nice change!

mindreader:
This worked beautifully for me.  Both with the normal default config, and with `runtime: false` and `FunWithFlags.Supervisor` in my tree.  Thank you!

tompave:
Cool! Thanks for confirming. I'm merging now and I'll release a new version soonish. There are a few small typespec-related tweaks I need to get in first.

PR #108, opened 2021-06-15
Adding config repo options

### LINKED ISSUE #94: Error in multi tenant apps with foreign key
Hey,

first of all, thanks for release this great library for us all. I think I've found an issue when using Fun With Flags with a multi tenant app. I've implemented my multi tenant model following this ecto guide: https://hexdocs.pm/ecto/multi-tenancy-with-foreign-keys.html.

The problem is when I try to enable/disable a flag, I got the error:

** (RuntimeError) expected org_id or skip_org_id to be set

Obviously, the error occurs because FWF doesn't know anything about my tenant model. I propose to add a `fun_with_flags: true` option for every call of `Ecto.Repo` inside the lib. By doing that, I can figure out which queries are originated from Fun With Flags and execute these queries outside my tenant model, like this:

```elixir
defmodule MyApp.Repo do
  use Ecto.Repo, otp_app: :my_app

  require Ecto.Query

  def prepare_query(operation, query, opts) do
    cond do
      opts[:skip_org_id] || opts[:schema_migration] || opts[:fun_with_flags] ->
        {query, opts}

      org_id = opts[:org_id] ->
        {Ecto.Query.where(query, org_id: ^org_id), opts}

      true ->
        raise "expected org_id or skip_org_id to be set"
    end
  end
end
```

I can work on that and send you a pull request. Do you think there's a chance for you to merge it?

gushonorato:
Closes #94

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # FunWithFlags
 
 [![Mix Tests](https://github.com/tompave/fun_with_flags/actions/workflows/test.yml/badge.svg?branch=master)](https://github.com/tompave/fun_with_flags/actions/workflows/test.yml?query=branch%3Amaster)
-[![Code Quality](https://github.com/tompave/fun_with_flags/actions/workflows/quality.yml/badge.svg?branch=master)](https://github.com/tompave/fun_with_flags/actions/workflows/quality.yml?query=branch%3Amaster)  
+[![Code Quality](https://github.com/tompave/fun_with_flags/actions/workflows/quality.yml/badge.svg?branch=master)](https://github.com/tompave/fun_with_flags/actions/workflows/quality.yml?query=branch%3Amaster)
 [![Hex.pm](https://img.shields.io/hexpm/v/fun_with_flags.svg)](https://hex.pm/packages/fun_with_flags)
 [![hexdocs.pm](https://img.shields.io/badge/docs-1.6.0-brightgreen.svg)](https://hexdocs.pm/fun_with_flags/)
 ![Hex.pm Downloads](https://img.shields.io/hexpm/dt/fun_with_flags)
@@ -581,6 +581,7 @@ config :my_app, MyApp.Repo,
 config :fun_with_flags, :persistence,
   adapter: FunWithFlags.Store.Persistent.Ecto,
   repo: MyApp.Repo,
+  repo_options: [prefix: "private"], # pass options to queries (optional)
   ecto_table_name: "your_table_name" # optional
   # Default table name is "fun_with_flags_toggles".
 ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,7 +1,7 @@
 # FunWithFlags
 
 [![Mix Tests](https://github.com/tompave/fun_with_flags/actions/workflows/test.yml/badge.svg?branch=master)](https://github.com/tompave/fun_with_flags/actions/workflows/test.yml?query=branch%3Amaster)
-[![Code Quality](https://github.com/tompave/fun_with_flags/actions/workflows/quality.yml/badge.svg?branch=master)](https://github.com/tompave/fun_with_flags/actions/workflows/quality.yml?query=branch%3Amaster)
```
tompave, line 4: Trailing whitespace is meaningful in Markdown, please revert this change.

README.md
```diff
@@ -593,7 +594,7 @@ The library comes with two PubSub adapters for the [`Redix`](https://hex.pm/pack
 
 The Redis PubSub adapter is the default and doesn't need to be explicitly configured. It can only be used in conjunction with the Redis persistence adapter however, and is not available when using Ecto for persistence. When used, it connects directly to the Redis instance used for persisting the flag data.
 
-The Phoenix PubSub adapter uses the high level API of `Phoenix.PubSub`, which means that under the hood it could use either its PG2 or Redis adapters, and this library doesn't need to know. It's provided as a convenient way to leverage distributed Erlang when using FunWithFlags in a Phoenix application, although it can be used independently (without the rest of the Phoenix framework) to add PubSub to Elixir apps running on Erlang clusters.
```
tompave, line 596: Trailing whitespace is meaningful in Markdown, please revert this change.

README.md
```diff
@@ -754,7 +755,7 @@ mix dialyzer
 
 ### Configuration changes have no effect in `MIX_ENV=dev`
 
-**Issue**: changing the library settings in the host application's Mix config file has no effect, or "missing process" exceptions are raised when booting. This should only be an issue in the development environment.
```
tompave, line 757: Trailing whitespace is meaningful in Markdown, please revert this change.

lib/fun_with_flags/config.ex
```diff
@@ -81,53 +79,49 @@ defmodule FunWithFlags.Config do
...
   end
 
-
   def ecto_repo do
     Keyword.get(persistence_config(), :repo)
   end
 
+  def ecto_repo_options do
+    Keyword.get(persistence_config(), :repo_options, [])
+  end
```
tompave, line 88: All other changes in this file seem to be stylistic. Please revert them.

tompave (commented):
Thank you for sending the PR.
I can see how this would help with https://github.com/tompave/fun_with_flags/issues/94, but this is not the solution agreed to in the issue.

I try to be conservative with the API of the package, including the config API. I'd rather not make it more complicated unless there is a valid reason. In this case, I'm not sure what the benefits of accepting and using arbitrary query options would be, sorry.

You can still change this and what was discussed in the issue.

gushonorato:
Hello,

actually, you are right. This is not the solution we agreed, but after some study, I decided that it would be better to not mix my application tables with third-party libraries on the same schema. This is the same approach used by Oban. I think it's a better solution and it doesn't add more complexity compared to the agreed solution. I'm using my fork with this PR in production for several months. It would be nice if you reconsider and merge this PR.

Thanks,
Gustavo

tompave:
@gushonorato if some form of default options is what you want, could you achieve it with this callback in your repo? https://hexdocs.pm/ecto/Ecto.Repo.html#c:default_options/1

I'm based this on the instructions from the Ecto guide you mentioned in the issue: https://hexdocs.pm/ecto/multi-tenancy-with-foreign-keys.html#setting-org_id-by-default

tompave:
I also believe that your for works, but I'm very hesitant to add support for a configuration option that injects arbitrary options into the Ecto queries performed by the package.

PR #105, opened 2021-04-06
Removing defdelegate from `FunWithFlags.Actor` protocol

kelvinst:
Fixes #104

PR #111, opened 2021-07-16
Fix runtime persistence config for releases, resolves #109

### LINKED ISSUE #109: Persistence config not recognized in release
# Versions 
erlang 23.1
elixir 1.11.3-otp-23
fun_with_flags 1.6.0

# Problem 

When running our application in a production release FWF doesn't seem to be accessing our configs/persistence adapter and is raising an exception when trying to call `FunWithFlags.all_flags()`.

Config: 

```
config :fun_with_flags, :persistence,
  adapter: FunWithFlags.Store.Persistent.Ecto,
  repo: MyApp.Repo
```

Exception raised when visiting the `/feature-flags/flags` endpoints of the UI: 

```
function FunWithFlags.Store.Persistent.Redis.all_flags/0 is undefined (module FunWithFlags.Store.Persistent.Redis is not available)
```

The application runs without issues in the local `dev` environment. I see in [1.5.1](https://github.com/tompave/fun_with_flags/blob/master/CHANGELOG.md#v151) there were some related changes and that there is an open [issue](https://github.com/tompave/fun_with_flags/issues/106) related to running the app in a production release but I haven't been able to identify the cause, any insight would be greatly appreciated.

Thanks!

connorlay:
`defdelegate/2` calls `persistence_adapter/0` at compile-time, causing system env configuration to be read prematurely and stored into the function definition itself. This commit replaces calls to `defdelegate/2` with functions that call `persistence_adapter/0` at runtime, fixing the issue.

I believe this resolves #109.

### DIFF
--- a/lib/fun_with_flags/simple_store.ex
+++ b/lib/fun_with_flags/simple_store.ex
@@ -10,10 +10,23 @@ defmodule FunWithFlags.SimpleStore do
     end
   end
 
-  defdelegate put(flag_name, gate), to: persistence_adapter()
-  defdelegate delete(flag_name, gate), to: persistence_adapter()
-  defdelegate delete(flag_name), to: persistence_adapter()
-  defdelegate all_flags(), to: persistence_adapter()
-  defdelegate all_flag_names(), to: persistence_adapter()
+  def put(flag_name, gate) do
+    persistence_adapter().put(flag_name, gate)
+  end
 
+  def delete(flag_name, gate) do
+    persistence_adapter().delete(flag_name, gate)
+  end
+
+  def delete(flag_name) do
+    persistence_adapter().delete(flag_name)
+  end
+
+  def all_flags do
+    persistence_adapter().all_flags()
+  end
+
+  def all_flag_names do
+    persistence_adapter().all_flag_names()
+  end
 end
--- a/lib/fun_with_flags/store.ex
+++ b/lib/fun_with_flags/store.ex
@@ -63,10 +63,13 @@ defmodule FunWithFlags.Store do
     |> cache_persistence_result()
   end
 
+  def all_flags do
+    persistence_adapter().all_flags()
+  end
 
-  defdelegate all_flags(), to: persistence_adapter()
-  defdelegate all_flag_names(), to: persistence_adapter()
-
+  def all_flag_names do
+    persistence_adapter().all_flag_names()
+  end
 
   defp cache_persistence_result(result) do
     case result do


durub:
Thanks for this PR, @connorlay. I was affected by this and was very lost on how fix it.

It would be great to have this fixed. I took a look and it looks good to me!

adamz-prescribefit:
Ran into this issue as well. Thanks for the fix! Are there any workarounds until this PR is merged?

k-cross:
@tompave elixir 1.13 is coming soon and it'd be awesome to get this in. Really love the library, thanks for all your work on it and thank you @connorlay for resolving this issue!

k-cross:
Hi @tompave, thanks for all the work you've done on this project. If you get the opportunity to review this before elixir 1.13 launches that'd be amazing!

tompave:
Hello, thank you for working on this, and apologies for the long wait.
I've been away from this project for some time, but I'll have a look soon.

In the meantime, the code changes in the PR look good. @connorlay I've just released `v.17.0` and pushed some changes to the Elixir and OTP versions in CI, plus bumped some dependencies in Mix.lock. Can you please rebase this on master and push, so that you can run the latest CI?

connorlay:
> Can you please rebase this on master and push, so that you can run the latest CI?

@tompave Done! It looks like I need maintainer approval to run the new CI jobs.

tompave:
Ok, I've caught up with the issue and this.

I think that this PR solves part of the problem, and I'm going to merge it. Thank you for working on this ❤️ and apologies again for the long wait. :-(

When I say "_part_ of the problem" I mean that the store module is still compiled into a module attribute in [`lib/fun_with_flags.ex#L30`](https://github.com/tompave/fun_with_flags/blob/v1.7.0/lib/fun_with_flags.ex#L30). Admittedly that is just the module name, which won't change unless the cache is disabled. By default, without custom config, the cache is enabled so most people will not encounter the issue.

There are also another couple of `defdelegate` that need to be updated:
https://github.com/tompave/fun_with_flags/blob/v1.7.0/lib/fun_with_flags.ex#L439-L459
I'll do it after merging this.

One final comment. @dvic and @connorlay, on using `Application.compile_env/3`, I've replied in https://github.com/tompave/fun_with_flags/issues/109#issuecomment-969344817. The relevant part:

> The suggestion to use [`Application.compile_env/3`](https://hexdocs.pm/elixir/1.12.3/Application.html#compile_env/3) is something I considered in the past to address the problem. Unfortunately, since that function first appeared in Elixir 1.10, I can't adopt it and still keep backward compatibility with older Elixirs. The good news is that FWF v1.7.0 (released yesterday) bumps the supported range to Elixir 1.9-1.12 (and drops 1.8), and FWF v1.8.0 will bump the supported range to Elixir 1.10-1.13, which means that I'll be able to update the config to use that new helper.

tompave:
@marcandre and @k-cross: why do you think that this is a blocker for Elixir 1.13? It doesn't seem to be required in my tests with v1.13-rc0.

marcandre:
The `defdelegate` no longer compiled in 1.13:

```
== Compilation error in file lib/fun_with_flags/protocols/actor.ex ==
** (CompileError) lib/fun_with_flags/protocols/actor.ex:117: undefined function defdelegate/2 (there is no such import)

could not compile dependency :fun_with_flags, "mix compile" failed. Errors may have been logged above. You can recompile this dependency with "mix deps.compile fun_with_flags", update it with "mix deps.update fun_with_flags" or clean it with "mix deps.clean fun_with_flags"
```

I presume this is because of ["Add defdelegate to the list of unallowed macros inside protocols as protocols do not allow function definitions"](https://github.com/elixir-lang/elixir/releases/tag/v1.13.0-rc.0)

Latest release works.

tompave:
@marcandre Ah, I see.

That was actually already fixed on master (but yet unreleased). It was fixed with https://github.com/tompave/fun_with_flags/pull/105.

On the other hand, the current PR (111) removes other `defdelegate` calls that were causing config+compile issues, but were not incompatible with Elixir 1.13 per se.

marcandre:
Ah, then my mistake, sorry 😅

Issue #123, opened 2022-01-11
Changing `ecto_table_name` config requires recompile


micahbf:
The `ecto_table_name` config is evaluated at compile time here: https://github.com/tompave/fun_with_flags/blob/master/lib/fun_with_flags/store/persistent/ecto/record.ex#L11

However this uses `Application.get_env` in `Config.ecto_table_name` by way of `Config.persistence_config`.

The result is that changing the `:fun_with_flags, :persistence, :ecto_table_name` config has no effect unless FunWithFlags is recompiled.

I would expect FunWithFlags to reflect the config changes, as it does with the rest of the persistence settings.

I found some info [here](https://stackoverflow.com/questions/40687186/how-can-you-dynamically-set-a-schema-table-name-for-a-ecto-model) that shows how to set a table name at runtime for an existing schema (with `Ecto.put_meta/2`).

If this sounds right, I could put together a PR.

tompave:
Thanks for opening this issue.
I think that this is something that should have been fixed when I worked on this:
https://github.com/tompave/fun_with_flags/blob/2b97162ebef2a40f09fb56a22dd9093a00a860ba/lib/fun_with_flags/config.ex#L67
`Application.compile_env` is probably the way to go in this case.

If you want to sent a PR that adopts that pattern, I'd be happy to accept it.

tompave:
Basically this:
```elixir
  schema Config.ecto_table_name() do
```

should possibly become:

```elixir
  schema Config.ecto_table_name_determined_at_compile_time() do
```

tompave:
This was addressed with https://github.com/tompave/fun_with_flags/pull/130.

micahbf:
Awesome! Sorry I didn't follow up myself.

PR #129, opened 2022-03-14
 Allow Ecto.UUID PKs

vinniefranco:
Adds a compile-time config of `:ecto_primary_key_type` allowing use optional use of `Ecto.UUID`

### DIFF
--- a/lib/fun_with_flags/config.ex
+++ b/lib/fun_with_flags/config.ex
@@ -19,7 +19,8 @@ defmodule FunWithFlags.Config do
   @default_persistence_config [
     adapter: FunWithFlags.Store.Persistent.Redis,
     repo: FunWithFlags.NullEctoRepo,
-    ecto_table_name: "fun_with_flags_toggles"
+    ecto_table_name: "fun_with_flags_toggles",
+    ecto_primary_key_type: :id
   ]
 
   def redis_config do
@@ -35,6 +36,11 @@ defmodule FunWithFlags.Config do
 
   def ecto_table_name do
     Keyword.get(persistence_config(), :ecto_table_name)
+  end
+
+
+  def ecto_primary_key_type do
+    Keyword.get(persistence_config(), :ecto_primary_key_type)
   end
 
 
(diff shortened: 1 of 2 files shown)


tompave:
Thank you for opening the PR.

Can you explain why you think this is required? In other words, why do you need a UUID PK in the FWF table?

vinniefranco:
Hi @tompave 

Great question! So, the reason I need UUID-based PKs is due to the way our production environment is set up. Int based pk/fk are a no-go

tompave:
I see, that makes sense, but I'm not sure this is the right way to go about it.
There is an open issue for how config is referenced in the ecto file (https://github.com/tompave/fun_with_flags/issues/123), and this change would suffer from the same issue.

I'd be happy to accept a PR to fix the problem described in that issue, and then a second PR to add this new functionality using the correct compile-time config approach.

vinniefranco:
> I see, that makes sense, but I'm not sure this is the right way to go about it. There is an open issue for how config is referenced in the ecto record file (#123), and this change would suffer from the same issue.
> 
> I'd be happy to accept a PR to fix the problem described in that issue, and then a second PR to add this new functionality using the correct compile-time config approach.

Yeah, I'm not sure that's the correct approach. `put_meta` is for already built structs, so one could dynamically change the source but it appears one cannot dynamically change the primary key definitions. Especially when primary key is configured using a module attribute, which is a compile-time thing.

tompave:
> Yeah, I'm not sure that's the correct approach. 

Which one? The one in your PR, or what's discussed in that issue?

>  `put_meta` is for already built structs,

Yes, I don't think the suggestion in the opening message of that issue is the right one. I'm suggesting a different approach in my replies.

> It would be better to name and document what is explicitly compile-time and run-time, as I've run into issues configuring FWF between ct configs/<env.exs> and the rt release.exs with Mix.
> I'm happy to help document the potential gotchas if need be

There were some gotchas, but they (or most of them) have been solved with the adoption of `Application.compile_env`. If you've run into issues configuring FWF in different envs I'd like to know if it was with an older version or with the latest. Chances are that you wouldn't have issues in the latest.

tompave:
With regard to my message above (https://github.com/tompave/fun_with_flags/pull/129#issuecomment-1067934868), that has now been fixed with https://github.com/tompave/fun_with_flags/pull/130. If you still want to update this PR you can adopt that pattern.

tompave:
Closing because of inactivity.

PR #139, opened 2022-07-28
Update option A documentation with needed changes to releases section

### LINKED ISSUE #133: The module `FunWithFlags.Supervisor` was given as a child to a supervisor but it does not exist
I run FunWithFlags with `runtime: false` in my `mix.exs` so I can start it manually in my supervision tree because I use `Phoenix.PubSub`.

I have added `FunWithFlags.Supervisor` to my supervision tree and it works just fine in dev. In prod however, I get `The module FunWithFlags.Supervisor was given as a child to a supervisor but it does not exist` and I am not sure why.

I am on version 1.8.1 of the library and I use `fun_with_flags_ui` 0.8 which is also configured with `runtime: false`.

What might be the issue?

zaid:
Update the *Application Start Behaviour* section in the `README` with the needed changes to the `releases` configuration when choosing **Option A**.

Resolves #133

### DIFF
--- a/README.md
+++ b/README.md
@@ -691,7 +691,7 @@ Then it's necessary to configure the Mix project to not start the `:fun_with_fla
 + {:fun_with_flags, "~> 1.6", runtime: false},
 ```
 
-You'll also need to modify the `releases` section in `mix.exs` so that it loads the `fun_with_flags` application.
+You'll also need to modify the `releases` section in `mix.exs` so that it loads the `fun_with_flags` application explicitly (since `runtime: false` / `app: false` will exclude it from the assembled release).
 
 ```diff
 def project do
--- a/README.md
+++ b/README.md
@@ -691,7 +691,7 @@ Then it's necessary to configure the Mix project to not start the `:fun_with_fla
 + {:fun_with_flags, "~> 1.6", runtime: false},
 ```
 
-You'll also need to modify the `releases` section in `mix.exs` so that it loads the `fun_with_flags` application explicitly (since `runtime: false` / `app: false` will exclude it from the assembled release).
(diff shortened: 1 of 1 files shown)


tompave (commented):
Thank you for this!

A couple of questions:
1. Adding the `releases: [ ... ]` key to the `project()`: is it required for both `runtime: false` and `app: false` options? It's not clear.
2. Is this not required at all with **Option B**?

zaid:
> Thank you for this!

@tompave my pleasure, thank you for your great work on this library! 

> 
> A couple of questions:
> 
> 1. Adding the `releases: [ ... ]` key to the `project()`: is it required for both `runtime: false` and `app: false` options? It's not clear.

Great question, I just checked and it seems that this is needed for `app: false` as well so I'll update the PR with the needed changes.

> 2. Is this not required at all with **Option B**?

I just checked and it looks like **Option B** doesn't require these changes.

tompave:
Awesome, thank you.

When you amend the PR, can you please specify that this is only required when using releases? i.e. not necessary if running with Mix?

zaid:
> When you amend the PR, can you please specify that this is only required when using releases? i.e. not necessary if running with Mix?

@tompave sure thing, I just updated the PR with the needed changes. Please let me know if I need to tweak this further.

nathanalderson:
It would be helpful to note that this is also needed for `fun_with_flags_ui` if that is being used, especially since that README [references](https://github.com/tompave/fun_with_flags_ui/blob/master/README.md?plain=1#L119) this one.

tompave (approved):

tompave:
Thank you for this work!
@nathanalderson I think that's fine. At the end of the section the Readme mentions that FWF.UI also needs to be configured differently.

PR #143, opened 2022-09-15
Add a custom query option to Ecto queries, closes #94

### LINKED ISSUE #94: Error in multi tenant apps with foreign key

SteffenDE:
When having a setup similar to what's described in the official Ecto guides on setting up multi tenancy with foreign keys, the queries originating from FunWithFlags can fail, as the tenant key is not set. See https://hexdocs.pm/ecto/multi-tenancy-with-foreign-keys.html.

This PR adds a custom query option (`:fun_with_flags`) to allow an easy detection of these queries and to skip the tenant key check. The test repo was changed to check for this option and raise if it is not set.

Closes #94.

### DIFF
--- a/README.md
+++ b/README.md
@@ -597,7 +597,7 @@ config :fun_with_flags, :persistence,
 
 It's also necessary to create the DB table that will hold the feature flag data. To do that, [create a new migration](https://hexdocs.pm/ecto_sql/Mix.Tasks.Ecto.Gen.Migration.html) in your project and copy the contents of [the provided migration file](https://github.com/tompave/fun_with_flags/blob/master/priv/ecto_repo/migrations/00000000000000_create_feature_flags_table.exs). Then [run the migration](https://hexdocs.pm/ecto_sql/Mix.Tasks.Ecto.Migrate.html).
 
-If you followed the Ecto guide on setting up [multi-tenancy with foreign keys](https://hexdocs.pm/ecto/multi-tenancy-with-foreign-keys.html), you must add an exemption for queries originating from FunWithFlags. These queries have a custom query option named `:fun_with_flags` set to `true`:
+If you followed the Ecto guide on setting up [multi-tenancy with foreign keys](https://hexdocs.pm/ecto/multi-tenancy-with-foreign-keys.html), you must add an exception for queries originating from FunWithFlags. These queries have a custom query option named `:fun_with_flags` set to `true`:
 
 ```elixir
 # Sample code, only relevant if you followed the Ecto guide on multi tenancy with foreign keys
--- a/README.md
+++ b/README.md
@@ -597,6 +597,32 @@ config :fun_with_flags, :persistence,
 
 It's also necessary to create the DB table that will hold the feature flag data. To do that, [create a new migration](https://hexdocs.pm/ecto_sql/Mix.Tasks.Ecto.Gen.Migration.html) in your project and copy the contents of [the provided migration file](https://github.com/tompave/fun_with_flags/blob/master/priv/ecto_repo/migrations/00000000000000_create_feature_flags_table.exs). Then [run the migration](https://hexdocs.pm/ecto_sql/Mix.Tasks.Ecto.Migrate.html).
 
+If you followed the Ecto guide on setting up [multi-tenancy with foreign keys](https://hexdocs.pm/ecto/multi-tenancy-with-foreign-keys.html), you must add an exemption for queries originating from FunWithFlags. These queries have a custom query option named `:fun_with_flags` set to `true`:
+
+```elixir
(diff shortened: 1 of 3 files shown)


SteffenDE:
Oh, I just saw that there already was a similar issue and PR. I'll check that and close this PR if necessary.

SteffenDE:
I think this PR is fine and actually implements the solution that was agreed on in #94.

dev_support/ecto/repo.ex
```diff
@@ -12,5 +12,23 @@ if FunWithFlags.Config.persist_in_ecto? do
...
+    @impl true
+    def prepare_query(_operation, query, opts) do
+      cond do
+        opts[:schema_migration] || opts[:fun_with_flags] ->
+          {query, opts}
+
+        true ->
+          raise "expected fun_with_flags query option to be set"
+      end
+    end
```
tompave, line 32: I wouldn't add this here.

README.md
```diff
@@ -597,6 +597,32 @@ config :fun_with_flags, :persistence,
 
 It's also necessary to create the DB table that will hold the feature flag data. To do that, [create a new migration](https://hexdocs.pm/ecto_sql/Mix.Tasks.Ecto.Gen.Migration.html) in your project and copy the contents of [the provided migration file](https://github.com/tompave/fun_with_flags/blob/master/priv/ecto_repo/migrations/00000000000000_create_feature_flags_table.exs). Then [run the migration](https://hexdocs.pm/ecto_sql/Mix.Tasks.Ecto.Migrate.html).
 
+If you followed the Ecto guide on setting up [multi-tenancy with foreign keys](https://hexdocs.pm/ecto/multi-tenancy-with-foreign-keys.html), you must add an exemption for queries originating from FunWithFlags. These queries have a custom query option named `:fun_with_flags` set to `true`:
```
tompave, line 600:
```suggestion
If you followed the Ecto guide on setting up [multi-tenancy with foreign keys](https://hexdocs.pm/ecto/multi-tenancy-with-foreign-keys.html), you must add an exception for queries originating from FunWithFlags. These queries have a custom query option named `:fun_with_flags` set to `true`:
```

tompave (commented):
Thank you for this. I think it makes sense. Just a couple of comments.

dev_support/ecto/repo.ex
tompave: I mean, it's convenient to have it here to inspect the options, but maybe you can comment it out.

SteffenDE: So the reason I added this is the following: the commit introduces a custom ecto query option that should be set for all queries originating from the library. Instead of adding a test for all calls, I added this to instead raise if a test executes a query that does not has the flag set. Therefore any code changes that forget to set the query option will lead to failing tests.

SteffenDE: So just to follow up: do you still want this to be changed? If yes, do you want the whole `prepare_query/3` removed?
I think it should at least log a warning if an unexpected query executes without the query option, but raising and therefore failing tests is still the solution I like the most. What do you think?

tompave: Yes, sorry. What you wrote makes sense. Let's keep this.

tompave (approved):
Thank you for this!

tompave:
For reference, I've tidied up the documentation to make it clearer how this works:
https://github.com/tompave/fun_with_flags/commit/00cd00a6eeb746994649f2e828f93cbb7bd40470

I find the fact that `prepare_query/3` only support the Query API a bit confusing, so I'd like to explain its gotchas.

tompave:
This has been released on Hex with version `1.10.0`.

PR #147, opened 2022-10-24
Add another case for Redis initialization

iamvery:
# Problem
An initialization case for Redix was missed in #145. This is unfortunate and probably justifies figuring out how to get a set of tests around the different configuration forms.

This was discovered by using v1.10.0 in a production app and receiving the error:

```
** (Mix) Could not start application fun_with_flags: FunWithFlags.Application.start(:normal, []) returned an error: shutdown: failed to start child: FunWithFlags.Notifications.Redis
    ** (EXIT) an exception was raised:
        ** (CaseClauseError) no case clause matching: {"redis://localhost", [socket_opts: [:inet6]]}
            (fun_with_flags 1.10.0) lib/fun_with_flags/notifications/redis.ex:73: FunWithFlags.Notifications.Redis.init/1
            (stdlib 4.1) gen_server.erl:851: :gen_server.init_it/2
            (stdlib 4.1) gen_server.erl:814: :gen_server.init_it/6
            (stdlib 4.1) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
```

# Solution
- Add new case
- Add tests (???)

### DIFF
--- a/test/fun_with_flags/notifications/redis_test.exs
+++ b/test/fun_with_flags/notifications/redis_test.exs
@@ -91,7 +91,7 @@ defmodule FunWithFlags.Notifications.RedisTest do
       assert {:ok, _pid} = NotifiRedis.publish_change(name)
 
       payload = "#{u_id}:#{to_string(name)}"
-      
+
       receive do
         {:redix_pubsub, ^receiver, ^ref, :message, %{channel: ^channel, payload: ^payload}} -> :ok
       after
@@ -150,7 +150,7 @@ defmodule FunWithFlags.Notifications.RedisTest do
 
     test "when the message is not valid, it is ignored" do
       channel = "fun_with_flags_changes"
-      
+
       with_mock(Store, [:passthrough], []) do
         Redix.command(PersiRedis, ["PUBLISH", channel, "foobar"])
         :timer.sleep(30)
@@ -163,7 +163,7 @@ defmodule FunWithFlags.Notifications.RedisTest do
       u_id = NotifiRedis.unique_id()
       channel = "fun_with_flags_changes"
       message = "#{u_id}:foobar"
-      
+
       with_mock(Store, [:passthrough], []) do
         Redix.command(PersiRedis, ["PUBLISH", channel, message])
         :timer.sleep(30)
@@ -178,7 +178,7 @@ defmodule FunWithFlags.Notifications.RedisTest do
 
       channel = "fun_with_flags_changes"
       message = "#{another_u_id}:foobar"
-      
+
       with_mock(Store, [:passthrough], []) do
         Redix.command(PersiRedis, ["PUBLISH", channel, message])
         :timer.sleep(30)
@@ -216,7 +216,7 @@ defmodule FunWithFlags.Notifications.RedisTest do
 
     test "when the message is not valid, the Cached value is not changed", %{name: name, cached_flag: cached_flag} do
       channel = "fun_with_flags_changes"
-      
+
       Redix.command(PersiRedis, ["PUBLISH", channel, to_string(name)])
       :timer.sleep(30)
       assert {:ok, ^cached_flag} = Cache.get(name)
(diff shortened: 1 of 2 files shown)


iamvery:
@tompave FYI, still thinking about the tests

iamvery:
I've had some difficulty getting integrated test cases on the configuration. I'm going to pause efforts for now (here's the [WIP](https://github.com/tompave/fun_with_flags/compare/master...iamvery:fun_with_flags:iamvery/wip)). Fwiw, I did confirm the changes in this PR w/ my app by using the local source code.

tompave:
Ah, good catch. This is something we missed. We should get it out with `v1.10.1`.

For the tests, I think we can do it in a different way. What you wrote in https://github.com/tompave/fun_with_flags/pull/145#discussion_r996093617 on the lack of coverage on the `Store.Persistent.Redis.worker_spec/0` function reminded me that I could just test the output of `worker_spec/0` directly, to make sure it would return the correct Child Spec. I've done it in https://github.com/tompave/fun_with_flags/pull/146.

So here we can do something similar. It requires a bit of refactoring though, because while the `Notifications.Redis` module has a `worker_spec/0` function, this doesn't really handle the Redis connection configuration. As you've seen, that's in `init/1`. But `worker_spec/0`, `start_link/*` and `init/*` are all connected, so we can move things around.

I think that this would do it: https://github.com/tompave/fun_with_flags/pull/148.
You could base your work on top of that: change `Notifications.Redis.worker_spec/0` and add a test for it.

iamvery:
Sorry for my delayed response. Some time-sensitive things have come up. I am hoping to have a look at this soon; perhaps early next week.

test/fun_with_flags/notifications/redis_test.exs
```diff
@@ -7,6 +7,91 @@ defmodule FunWithFlags.Notifications.RedisTest do
...
+          :start_link,
+          [
+            {
+              url,
+              [
+                socket_opts: [:inet6],
+                name: :fun_with_flags_notifications,
+                sync_connect: false
+              ]
+            }
```
iamvery, line 58: ❓ Is this extra wrapping as a tuple OK? The shape of this is slightly different than Persistance.Redis.worker_spec/0. Maybe that's OK since that function is based on Redix.child_spec/1?

tompave:
Ops, sorry, I merged the other PR this is based on and deleted the branch, thinking it would auto-switch this PR's base branch to `master`. Instead it has closed it. It seems that I cannot re-open it myself.

test/fun_with_flags/notifications/redis_test.exs
tompave: Yes, that's right. It just happens that `Redix` wants its arguments in a way and this FWF's supervised worker wants its arguments in a different way.

Here the format is:
```
start: {
  Module.That.Receives.The.Function.Call,
  :the_name_of_the_function,
  [
    "a list that contains"
    "the",
    {"positional", "arguments", "to", "the function"},
    :they,
    [
      could: "be",
      anything: "really"
    ]
  ]
}
```

The name of the function and the arguments inside the outermost `[]` are application specific. It's just customary to use `start_link` because that's the convention.

For example FWF still works with this diff:

```diff
diff --git a/lib/fun_with_flags/notifications/redis.ex b/lib/fun_with_flags/notifications/redis.ex
index 3f4c54a..aee4930 100644
--- a/lib/fun_with_flags/notifications/redis.ex
+++ b/lib/fun_with_flags/notifications/redis.ex
@@ -27,7 +27,7 @@ defmodule FunWithFlags.Notifications.Redis do

     %{
       id: __MODULE__,
-      start: {__MODULE__, :start_link, [redis_conn_config]},
+      start: {__MODULE__, :foobar, [:something_else, redis_conn_config]},
       restart: :permanent,
       type: :worker,
     }
@@ -39,7 +39,7 @@ defmodule FunWithFlags.Notifications.Redis do
   # used to build the outgoing notification payloads and to ignore
   # incoming messages that originated from this node.
   #
-  def start_link(redis_conn_config) do
+  def foobar(:something_else, redis_conn_config) do
     GenServer.start_link(__MODULE__, {redis_conn_config, Config.build_unique_id}, [name: __MODULE__])
   end
```

tompave (commented):
LGTM, thank you for this. Please re-open it and I'll merge it.

iamvery:
I am also not able to reopen, so I rebased the commits on the latest master and opened #149

PR #151, opened 2022-11-28
Add support for SQLite

tylerbarker:
# Overview

Hello 👋 thanks for this library!

I'm currently working on a project using SQLite via the [`ecto_sqlite3` adapter](https://hexdocs.pm/ecto_sqlite3). I was integrating FunWithFlags and found that only Postgres and MySQL were supported, so I thought I'd have a go at extending that. Let me know what you think.

~**Note:** I made some edits to `ecto_test.exs` as there were a couple of tests which were flaky even without these changes. My understanding is these updates (sorting for list equality) don't relate to the underlying functionality and so the tests continue to assert the same thing.~


tompave:
Thank you for using the library and for opening this PR.
A few comments before doing a deeper review:

1. I would remove https://github.com/tompave/fun_with_flags/pull/151/commits/f375eb9af8a9f7a2b413e8a11e85a6b21bcf8257, which fixes the flaky tests, from the PR. I thank you for the fix and I would be happy to accept it as a separate PR.
2. I can see a lot of extra formatting changes in some some files (trailing commas, parenthesis, etc). Please revert them. One day perhaps I'll pass this whole project through the formatter, but a feature PR is not the right place for it.
3. I can see that you've already bumped the package version in separate files. Please revert that. The release process for this package is based on bumping all those references in an isolated commit that gets git-tagged. I'll do it when it's time. This includes the changelog. You can add a new heading `Unreleased`.
4. I can see some serious refactoring of a DB migrations. I'm not sure about those changes. First off, that PR was only necessary to upgrade the table in applications that were using this package before a certain change. I don't think it's really relevant for SQLite, so I would revert those changes. Even if it were necessary, if you drop and re-add a column you'll be dropping the data in the table I think, which is a no-no.

tylerbarker:
Hey, thanks for the detailed feedback. I think I've addressed your comments now. I'll raise the test changes in a separate PR like you suggest.

tylerbarker:
Looks like the tests are failing due to the migration being incompatible with SQLite. I wonder would it be worth re-introducing my migration changes, but configured alongside the original, so that the `remove` / `add` version only runs when `System.get_env("RDBMS")` is set to `"sqlite3"`?

lib/fun_with_flags/store/persistent/ecto.ex
```diff
@@ -134,6 +135,25 @@ defmodule FunWithFlags.Store.Persistent.Ecto do
...
+          repo.rollback("Exception: #{inspect(e)}")
+      else
+        {:error, reason} ->
+          repo.rollback("Error while upserting the gate: #{inspect(reason)}")
+
+        {:ok, value} ->
+          {:ok, value}
+      end
+    end)
+  end
```
tompave, line 155: I want to make sure I understand this.
Looking at [the SQLite transaction docs](https://www.sqlite.org/lang_transaction.html), a write transaction will automatically block other concurrent write transactions. Which makes sense because SQLite doesn't have MVCC:

> Transactions can be started manually using the BEGIN command. Such transactions usually persist until the next COMMIT or ROLLBACK command. But a transaction will also ROLLBACK if the database is closed or if an error occurs (...)
> SQLite supports multiple simultaneous read transactions coming from separate database connections, possibly in separate threads or processes, but only one simultaneous write transaction.
> A read transaction is used for reading only. A write transaction allows both reading and writing. A read transaction is started by a SELECT statement, and a write transaction is started by statements like CREATE, DELETE, DROP, INSERT, or UPDATE (collectively "write statements"). **If a write statement occurs while a read transaction is active, then the read transaction is upgraded to a write transaction** if possible. (...)
> (emphasis added)

The bold sentence would be relevant here because this is used to wrap a SELECT followed by an UPDATE.

And then:

> If some other database connection has already modified the database or is already in the process of modifying the database, then upgrading to a write transaction is not possible and the write statement will fail with [SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy).

This is the important part for us, because that's the whole point of this transaction in Postgres and MySQL.

tompave: Is the above correct?

mix.exs
```diff
@@ -60,6 +60,7 @@ defmodule FunWithFlags.Mixfile do
     [
       {:redix, "~> 1.0", optional: true},
       {:ecto_sql, "~> 3.0", optional: true},
+      {:ecto_sqlite3, "~> 0.8.2", optional: true, only: [:dev, :test]},
```
tompave, line 63: I think that just the minor version should be fine. It would be consistent with the others.
```suggestion
      {:ecto_sqlite3, "~> 0.8", optional: true, only: [:dev, :test]},
```

mix.exs
```diff
@@ -85,7 +86,8 @@ defmodule FunWithFlags.Mixfile do
       {:"test.phx", [&run_tests__redis_pers__phoenix_pubsub/1]},
       {:"test.ecto.postgres", [&run_tests__ecto_pers_postgres__phoenix_pubsub/1]},
       {:"test.ecto.mysql", [&run_tests__ecto_pers_mysql__phoenix_pubsub/1]},
-      {:"test.redis", [&run_tests__redis_pers__redis_pubsub/1]},
+      {:"test.ecto.sqlite3", [&run_tests__ecto_pers_sqlite3__phoenix_pubsub/1]},
+      {:"test.redis", [&run_tests__redis_pers__redis_pubsub/1]}
```
tompave, line 90: Let's keep the trailing comma. GitHub doesn't support clean suggestions in this case, but basically:

```suggestion
      {:"test.redis", [&run_tests__redis_pers__redis_pubsub/1]},
```

mix.exs
```diff
@@ -98,6 +100,7 @@ defmodule FunWithFlags.Mixfile do
       &run_tests__redis_pers__phoenix_pubsub/1, &run_integration_tests__redis_pers__phoenix_pubsub__no_cache/1,
       &run_tests__ecto_pers_postgres__phoenix_pubsub/1, &run_integration_tests__ecto_pers_postgres__phoenix_pubsub__no_cache/1,
       &run_tests__ecto_pers_mysql__phoenix_pubsub/1, &run_integration_tests__ecto_pers_mysql__phoenix_pubsub__no_cache/1,
+      &run_tests__ecto_pers_sqlite3__phoenix_pubsub/1, &run_integration_tests__ecto_pers_sqlite3__phoenix_pubsub__no_cache/1
```
tompave, line 103: Let's keep the trailing comma here too.

```suggestion
      &run_tests__ecto_pers_sqlite3__phoenix_pubsub/1, &run_integration_tests__ecto_pers_sqlite3__phoenix_pubsub__no_cache/1,
```

mix.exs
```diff
@@ -197,6 +200,20 @@ defmodule FunWithFlags.Mixfile do
...
+    Mix.shell.cmd(
+      "mix test --color --force --exclude redis_pubsub --exclude redis_persistence #{arg}",
+      env: [
+        {"CACHE_ENABLED", "true"},
+        {"PUBSUB_BROKER", "phoenix_pubsub"},
+        {"PERSISTENCE", "ecto"},
+        {"RDBMS", "sqlite3"},
+      ]
+    )
+  end
```
tompave, line 215: To keep the functions in a consistent order, I'd move this below `run_tests__ecto_pers_mysql__phoenix_pubsub`.

tompave (commented):
Thank you, this seems like a solid change. I have a few style comments, but looks good otherwise.

One request: let's just make it `sqlite` rather than `sqlite3` everywhere (except the package name). It's simpler, and I think that in 2022 `sqlite == sqlite3`.

Out of curiosity: where are you using this package with SQLite? Is it in an embedded application with something like Nerves?

tompave:
> Looks like the tests are failing due to the migration being incompatible with SQLite. I wonder would it be worth re-introducing my migration changes, but configured alongside the original, so that the `remove` / `add` version only runs when `System.get_env("RDBMS")` is set to `"sqlite3"`?

I see.
I don't remember exactly what you were changing in the migration (I didn't have the time to do a deep review), and now I can't access it anymore because you git-amended and force-pushed, and I can't get the SHA of that first commit. (Going forward, let's just push new commits without force-pushes please; amending and force-pushing nukes the history and makes re-reviews more difficult.)

The error seems to come from [the second migration](https://github.com/tompave/fun_with_flags/blob/16c569fd164341abde185e929bfaf9c02e1db24d/priv/ecto_repo/migrations/00000000000001_ensure_columns_are_not_null.exs), which is really just [an upgrade step for v1.1.0](https://github.com/tompave/fun_with_flags/blob/master/CHANGELOG.md#v110). I remember thinking that maybe it's time to remove that second migration from master, since it's been a while and it can still be linked in the changelog. I'll do that tomorrow or next week, then you can merge master into this PR.

lib/fun_with_flags/store/persistent/ecto.ex
tylerbarker: Yeah what you've written a more articulate version of my understanding. I wasn't sure whether it was still necessary so I opted to follow the pattern of the others in case. Do you think I should remove the extra `repo.transaction` there?

tylerbarker:
Understood, good point on sqlite3 -> sqlite 👍🏼 I'll keep an eye out for your second migration changes, cheers.

tylerbarker:
I'm using it in my side project [Prototape](https://prototape.io) - which is a LiveView application. I read [an interesting article](https://fly.io/blog/all-in-on-sqlite-litestream/) from Fly about the potential for SQLite in full-stack apps, so I'm giving it a go.

tylerbarker:
Hey @tompave, any updates on this?

simoncocking:
Just a quick note to say we're super excited by the prospect of Sqlite support in FunWithFlags - we're eyeing a move from Postgres to Sqlite in our app. Oban and FwF were the two roadblocks; Oban has since announced support. We'd _love_ to see some movement on this PR 😉 

P.S. @tylerbarker huge props for getting the ball rolling on this! 😁

tompave:
Hey there, apologies for disappearing. I'll be looking at the recent activity here this week.

lib/fun_with_flags/store/persistent/ecto.ex
tompave: Very late reply, but I think that we should keep the transaction block.

Everything should happen within a transaction in SQLite. The docs linked above say that transactions "can be started manually using the BEGIN command", which implies that it's not the only way to start them. Later the docs say that a simple `SELECT` starts a read transaction automatically.

Here we want to avoid race conditions between concurrent updates to the percentage gates (because we can't rely on the unique index, like we do for the other gates [1]).

I think that in the case of concurrent `SELECT+UPDATE`s, without transaction block the 4 operations (2 SELECTs, followed by 2 UPDATEs) could be interleaved in ways that could lead to writing stale data. So keeping the transaction block is necessary.

[1]: In hindsight this could be solved with a different table schema and a better index, but it's too late to do that without introducing a major breaking change.

tompave (approved):
Thank you for all the work on this, and apologies for the long wait. Looks great!

PR #152, opened 2022-12-19
no longer start an application 

dougw-bc:
This PR removes the `application.ex` and also removes the mod key from the application config, so nothing is started automatically.

the fun with flags supervisor MUST be included in the consuming application.

Can include documentation updates if this will be considered for 2.0 to ensure it does not break existing deployments 

Relevant discussion: https://github.com/elixir-lang/elixir/issues/9683

### DIFF
--- a/lib/fun_with_flags/application.ex
+++ b/lib/fun_with_flags/application.ex
@@ -1,9 +1,0 @@
-defmodule FunWithFlags.Application do
-  @moduledoc false
-
-  use Application
-
-  def start(_type, _args) do
-    FunWithFlags.Supervisor.start_link(nil)
-  end
-end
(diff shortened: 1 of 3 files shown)


tompave:
Hey there, thank you for using the library and for working on this.
Yes, that's possibly something for 2.0, but I'm not looking for PRs on this area at the moment. Sorry.

nallwhy:
I believe this PR should be reconsidered. The current documentation at https://github.com/tompave/fun_with_flags?tab=readme-ov-file#application-start-behaviour does not work as expected, and even in option B, `fun_with_flags: :load` must be explicitly set in the release configuration.

Rather than requiring these additional steps, I think it would be better to ensure explicit execution.

tompave:
Hi @nallwhy, thank you for reaching out.

Can you please open an issue to report the problem? Easier to track it that way, rather than commenting on an old and closed PR.

PR #154, opened 2023-04-17
Add timestamps to Ecto DB table

PaulOstazeski:
This adds `inserted_at` and `updated_at` columns to the database table when using Ecto, and uses the built-in Ecto behavior to populate their values without any additional work. This makes it possible to directly measure how long a given flag has been in its current state, which in turn makes it easier to determine when a flag is safe to be removed (i.e. that it has been "on" for some/many/all users for "long enough" that we haven't observed any downsides).

### DIFF
--- a/priv/ecto_repo/migrations/00000000000002_add_timestamps.exs
+++ b/priv/ecto_repo/migrations/00000000000002_add_timestamps.exs
@@ -1,0 +1,32 @@
+defmodule FunWithFlags.Dev.EctoRepo.Migrations.EnsureColumnsAreNotNull do
+  use Ecto.Migration
+  #
+  # Use this migration to add the ecto `inserted_at` and `updated_at` columns
+  # to the table created using the `CreateFeatureFlagsTable` migration
+  # from versions `<= 1.10.0`.
+  #
+  # https://hexdocs.pm/ecto_sql/Ecto.Migration.html#timestamps/1
+  #
+  # If the table has been created with a migration from `>= 1.11.0`,
+  # then the `timestamp` columns are already there and there
+  # is no need to run this migration. In that case, this migration
+  # is a no-op.
+  #
+  # This migration assumes the default table name of "fun_with_flags_toggles"
+  # is being used. If you have overridden that via configuration, you should
+  # change this migration accordingly.
+
+  def up do
+    case repo.query("select inserted_at, updated_at from fun_with_flags_toggles where false") do
+      {:ok, %{num_rows: 0}} ->
+        # timestamp columns exist, do nothing
+        true
+
+      {:error, _} ->
+        # Assume error is from timestamp columns not existing, run migration
+        alter table(:fun_with_flags_toggles) do
+          timestamps()
+        end
+    end
+  end
+end
--- a/lib/fun_with_flags/store/persistent/ecto/record.ex
+++ b/lib/fun_with_flags/store/persistent/ecto/record.ex
@@ -13,6 +13,7 @@ defmodule FunWithFlags.Store.Persistent.Ecto.Record do
     field :gate_type, :string
     field :target, :string
     field :enabled, :boolean
+    timestamps()
   end
 
   @fields [:flag_name, :gate_type, :target, :enabled]
--- a/priv/ecto_repo/migrations/00000000000000_create_feature_flags_table.exs
+++ b/priv/ecto_repo/migrations/00000000000000_create_feature_flags_table.exs
@@ -12,6 +12,7 @@ defmodule FunWithFlags.Dev.EctoRepo.Migrations.CreateFeatureFlagsTable do
       add :gate_type, :string, null: false
       add :target, :string, null: false
       add :enabled, :boolean, null: false
+      timestamps()
     end
 
     create index(


priv/ecto_repo/migrations/00000000000002_add_timestamps.exs
```diff
@@ -0,0 +1,32 @@
+defmodule FunWithFlags.Dev.EctoRepo.Migrations.EnsureColumnsAreNotNull do
+  use Ecto.Migration
+  #
+  # Use this migration to add the ecto `inserted_at` and `updated_at` columns
+  # to the table created using the `CreateFeatureFlagsTable` migration
+  # from versions `<= 1.10.0`.
+  #
+  # https://hexdocs.pm/ecto_sql/Ecto.Migration.html#timestamps/1
+  #
+  # If the table has been created with a migration from `>= 1.11.0`,
```
PaulOstazeski, line 10: I am just guessing that this change would necessitate a minor version bump. Note that this PR **does not** actually change any version numbers.

izaxveta (approved):

ArgyleWerewolf:
To me, this is good.

alpardal (approved):

priv/ecto_repo/migrations/00000000000002_add_timestamps.exs
```diff
@@ -0,0 +1,32 @@
...
+  # change this migration accordingly.
+
+  def up do
+    case repo.query("select inserted_at, updated_at from fun_with_flags_toggles where false") do
+      {:ok, %{num_rows: 0}} ->
+        # timestamp columns exist, do nothing
+        true
+
+      {:error, _} ->
+        # Assume error is from timestamp columns not existing, run migration
```
tompave, line 26: This is very elegant. 👍

tompave (approved):
Thank you for using the package and for working on this. And apologies for the very late reply.

This change makes sense. Now that I think about it, this should have been present from the start.

tompave:
Hey @PaulOstazeski, I'm afraid I have to revert this. (revert PR: https://github.com/tompave/fun_with_flags/pull/160)

This weekend I've been catching up on some lover overdue maintenance work on this repo, and I think I merged this a bit too quickly. This was my mistake, and I apologize.

I'm also afraid that I forgot to run CI on this PR. Normally I have to approve GHA runs for PRs, but it usually presents me a big visible button on the PR page to do so. For some reasons it didn't for this PR. (It could be either because it's been too long since the PR was opened, or because in the meantime I updated the GHA workflow on the master branch.) I thought "sure, it seems safe, I can run it locally", and just merged it.

Unfortunately I came across a few issues after that. I think we need to address them before we can consider this to be merged. Normally I'd work on it myself and roll forward, but my time at the moment is limited, and I don't want to leave the master branch in a not-working state, sorry.

If you would like to give this another try, here are my observations after testing this locally:

1. Elixir 1.15 won't translate anymore `repo` to `repo()`, as it [raises a compiler error](https://github.com/tompave/fun_with_flags/actions/runs/5636873491/job/15269391344#step:9:21). I fixed it with https://github.com/tompave/fun_with_flags/pull/159 (later reverted).
2. The migration doesn't contain a `down` function and it cannot be rolled back. I fixed it with https://github.com/tompave/fun_with_flags/commit/a4d8e329b75a09aea3ecf23dd470bbf753d0005e in [my working branch](https://github.com/tompave/fun_with_flags/compare/95b168d...ecto_timestamps_tweaks).
3. On the incremental migration: I wonder if you have only tested it in MySQL. I could apply your migration locally on MySQL, but I couldn't run `mix ecto.migrate` on Postgres. I think that the issue is that once an error is raised within the migration (the very thing that I thought was elegant 🙃), then the migration transaction is flagged as aborted, and every subsequent SQL statements fail. In order for this to be merged, it needs to work for Postgres too.
4. Still on the incremental migration: I couldn't apply it on either MySQL or Postgres if the table already contained records. That's because Ecto's timestamps default to `NOT NULL` and the migration doesn't provide a default. There are two solutions for this:
    1. Provide a default value. However adding a column with a default value is a blocking operation on Postgres, and I believe it can be tricky on MySQL too unless some non-blocking DDL API is used. That's not a problem for small tables, but it can lock the table in production if there is too much data. I'd be very uncomfortable releasing a new minor version of the package where the upgrade instructions come with a warning that it might require a DB maintenance window. 😬 
    2. An alternative on the above is to do it in multiple steps, which is a common workaround in Postgres: add the column, add a default value (or backfill it), make it `NOT NULL` later. It's just more complex to rollout and to automate within the package.
    3. Make the timestamp columns nullable, so that the problem goes away. I was exploring this, but then I realized that I was still blocked by point 3 on Postgres, mentioned above.

On the last two points, I suppose that my main concern is: if I add this functionality to the package, then the upgrade path should be smooth and painless. If it requires too many manual steps, and it comes with risks, then I'd be more inclined to add it to the next major version, rather than a minor version bump.

I hope it makes sense!

PaulOstazeski:
Thanks @tompave, I definitely missed some subtleties here. I'll take another crack at it and see if I can address these.

PR #176, opened 2024-07-10
Add instrumentation using :telemetry

Kasse-Dembele:
This PR adds instrumentation to FunWithFlags using the `:telemetry` library. This allows users to gain insights into the performance and usage of feature flags in their applications.

## Key changes:
- Added telemetry events for all major operations (enable, disable, clear, enabled?)
- Each event includes duration measurements and relevant metadata
- Telemetry events are emitted using the `[:fun_with_flags, :flag_operation]` event name

## New telemetry events:
- `[:fun_with_flags, :flag_operation]`
  - Measurements: 
    - `duration`: Time taken for the operation in native time units
  - Metadata:
    - `flag_name`: Name of the flag being operated on
    - `operation`: Type of operation (:enable, :disable, :clear, or :enabled?)
    - `options`: Any options passed to the operation
    - `result`: Result of the operation

This change is backwards compatible and does not affect existing functionality. Users can choose to attach handlers to these events to collect metrics or perform logging.

This PR aims to provide better observability for applications using FunWithFlags, allowing for easier debugging, performance monitoring, and usage analytics.

## Testing:
- Added unit tests for telemetry event emission in all relevant operations
- Ensured existing tests pass without modification

## Next steps:
- Update documentation to include information about the new telemetry events

Feedback and suggestions are welcome!


tompave:
Hi, thank you for using the library and for the PR. This is some nice work!
I think I can accept this change, but first I want to measure the performance overhead.

In the meantime, can you please merge in `master` or rebase? I've updated the CI matrix and it'd be good to re-run CI with the latest Elixir and OTP versions.

I can also see that at the moment the telemetry tests are failing with Elixir 1.14 (is it incompatible?), but the good news is that I'm dropping support for Elixir 1.14 (it's removed from CI on master), so that won't be a concern. Still, I'd be curious to know why it's failing there.

mix.exs
```diff
@@ -80,7 +80,9 @@ defmodule FunWithFlags.Mixfile do
 
       {:ex_doc, "~> 0.21", only: :dev, runtime: false},
       {:credo, "~> 1.7", only: :dev, runtime: false},
-      {:dialyxir, "~> 1.0", only: :dev, runtime: false}
+      {:dialyxir, "~> 1.0", only: :dev, runtime: false},
+      {:telemetry_test, "~> 0.1", only: :test, runtime: false},
```
tompave, line 84: I'm not sure about this package. It looks like it's not really much used in the ecosystem: https://hex.pm/packages/telemetry_test

Is there any other way to test telemetry?

Kasse-Dembele: yes for sure,[]( https://hex.pm/packages/telemetry_test ) just abstract away the setup for telemetry event handler. But now i have updated the tests without using the telemetry_test package.

jmartinezpoq:
I love this idea, I was thinking of building an ad hoc solution to solve this problem

Is there anything pending to be able to add this to the library? 
Can we help in any way to make this a reality?
@tompave

tompave:
Telemetry events are being added with: https://github.com/tompave/fun_with_flags/pull/197.

PR #181, opened 2024-08-13
Remove unused comment within mix.exs file

DuldR:
Noticed this while reading the release commits. Thanks for all the hard work on this project!


tompave:
Thank you for using the library and for the cleanup!

Issue #125, opened 2022-01-20
[Question] Ensuring feature flags are in sync across dev environments


oliverswitzer:
Hi,

Do you have any suggestions for ensuring that everyone's development / test environments are in sync when folks add new feature flags? We'd love some form of a version controlled config file that at the very least indicates the presence of flags, and even better, the desired flag state / gate / actor, so that each developer doesn't have to know what flags to set in their DB.

Cheers,
Oliver

tompave:
Hey there, thank you for using the package and for asking this question!

The package doesn't support this sort of flag declaration file. I must admit that it sounds like a useful feature, and in principle I like the idea of a `flagfile` (just like a `Brewfile` for Homebrew and a `Gemfile` for Ruby). I'm just not certain about adding this sort of complexity to the package. For context, something similar was discussed in https://github.com/tompave/fun_with_flags/issues/119.

On to your question, I think that since it's only for dev/test then it should be easy. At least, there is no need to worry about race conditions when multiple application instances start (which I see as a source of complexity if I were to support this).

My suggestion is to have a module somewhere in your app where you programmatically define all your flags:

```elixir
defmodule LocalDevFlags do
  import FunWithFlags, only: [enable: 2, disable: 2, clear: 2]

  def reset() do
    clear :foo
    disable :foo
    enable :foo, for_actor: %MyApp.User{id: 42}
    enable :foo, for_actor: %MyApp.User{id: 43}

    clear :bar
    enable :bar
    disable :bar, for_group: "restricted_users"

    # More flag declarations...
  end
end
```

And then to call `LocalDevFlags.reset()` somewhere in your app initialization code.

This won't really prevent people from adding, deleting and changing flags, but it will bootstrap the set of flags in a standard way, automatically.

oliverswitzer:
@tompave totally makes sense, that's actually the direction I just started working towards.

I'm assuming there currently is no way through the `FunWithFlags` module to clear all feature flags, other than going directly to my apps Repo?

tompave:
You could do:
```elixir
FunWithFlags.all_flag_names()
|> elem(1)
|> Enum.each(&FunWithFlags.clear/1)
```

Issue #119, opened 2021-12-03
Configure set of permitted flags


jotaviobiondo:
First, congrats on the lib! I really like the features and it's being super useful :)

One thing that I was feeling the need, is to somehow configure a set of the available flags, something like this:

```elixir
config :fun_with_flags, 
  permitted_flags: [
    :feature_a,
    :feature_b,
    :feature_c,
  ]
```

I can see 2 advantages of having this:
1. When we are calling functions like `FunWithFlags.enabled?(:feature_a)` or `FunWithFlags.enable(:feature_a)`, we validate against the permitted flags. Useful to ensure there's not typo in the flag name, for example.
2. Listing the available flags in the UI even if they don't exist in the storage. This is because I always need to check the code before add them in through the UI, not a big deal, bit if they are already listed it would be a very nice thing :)

The 1st advantage could required a relatively big refactoring and for me is not that important. It's just a plus, I think.

The 2nd advantage I really like to have it, since it makes easier to work only with the UI. This could be implemented in 2 ways, I think:
1. Implementing in this repository, doing the logic inside `all_flag_names/0` and `all_flags/0`
2. Implementing in the https://github.com/tompave/fun_with_flags_ui repository.

I guess I could "override" the storage I'm using, creating my own adapter and doing these logics for the  `all_flag_names/0` and `all_flags/0`, but having this in the library can be easier for users :)

Do you think this is a valid feature to having in this library?

tompave:
Hello, thank you for using the package and for the kind words. :-)

I can see how what you suggest would be useful, but I'm not sure about adding it to the package. I'm a big fan of simplicity, and what you're suggesting basically requires a block on the creation of flags: they can only be seeded from a known list, and once the known flags are created their gates can be toggled or cleared, but new flags should be created dynamically. I'm not saying it wouldn't be useful, but it's a significantly different UX mode and I'm not sure about adding it to the package.

Still, I think that some or all of it can be implemented without package changes.

> When we are calling functions like `FunWithFlags.enabled?(:feature_a)` or `FunWithFlags.enable(:feature_a)`, we validate against the permitted flags. Useful to ensure there's not typo in the flag name, for example.

That makes a lot of sense, but it could be achieved with your own wrapper (I haven't tried to run it or check if it compiles):

```elixir
defmodule MyAppFlags do
  @known_flags [
    :feature_a,
    :feature_b,
    :feature_c,
  ]

  defmodule UnknownFlagName do
    defexception [:message]
  end

[...]

PR #182, opened 2024-09-24
Add ecto migration generator for new flags

brain-geek:
While our feature flags are defined in code, after deployment to new environment someone needs to manually create those with current approach. That is sometimes prone to errors like creating `feature-flag` instead of `Feature-flag` or `feature_flag`, which creates confusion during testing.

To simplify this process, I have added an optional generator to generate new (newly disabled) flag during migration.

It's heavily based on `ecto.gen.migration`, and should help if you have multiple environments and want to deploy more feature flags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # FunWithFlags
 
 [![Mix Tests](https://github.com/tompave/fun_with_flags/actions/workflows/test.yml/badge.svg?branch=master)](https://github.com/tompave/fun_with_flags/actions/workflows/test.yml?query=branch%3Amaster)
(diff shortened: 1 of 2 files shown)


tompave:
Hello, thank you for using the package and for the PR.

I understand your use case, but I'm not sure this is something I would like to add to the package.

I'm also not convinced that this should be done with a DB migration. In my opinion DB migrations are meant to manage the DB _schema_, not to manipulate data (with rate exceptions).

If users of the library have this use case, they'd be better off with a custom module/function that sets up all required flags (if they don't exist), and then invoke it when the app supervisor starts. For reference, something similar was discussed in https://github.com/tompave/fun_with_flags/issues/125 and https://github.com/tompave/fun_with_flags/issues/119.

For now, I thank you for the work and for the proposal, but I must decline the PR.

brain-geek:
@tompave Thanks a lot for the feedback and for explaining the reasoning behind it.

PR #184, opened 2024-11-18
add ecto schema prefix config support

diogomrts:


tompave:
Hi, thank you for using the library and for the PR.

> I'm not too familiar with the other DB engines so I'm not sure if these changes would work for all.

I don't think this will work as it is with DBs other than Postgres. I've allowed CI to run on the PR, and I can see the tests fail with both MySQL and SQLite. You can also verify this locally with `mix test.all`.

In general, this sort of customization can be achieved with a [custom persistence adapter](https://github.com/tompave/fun_with_flags?tab=readme-ov-file#custom-persistence-adapters). Would that solve your problem?

tompave:
Further question: shouldn't the postgres schema be defined in the Ecto Repo, at the connection level? I've done some googling and it looks like that's how people usually solve this. In that case, you should simply pass to FunWithFlags a properly configured Ecto Repo.

PR #190, opened 2025-01-30
Switch documented migration to use change

treble37:
Ecto SQL docs recommend using `change` https://hexdocs.pm/ecto_sql/3.12.1/Ecto.Migration.html#module-change


tompave:
Hi, thank you for the PR and for using the package.

Yes, I'm aware that `change` is available. I prefer to not change it at this point because users of the library have already copied this migration in the their projects, and I'd rather not change without good reason.

PR #203, opened 2025-08-18
Hierarchical flags

bernardd:
This is an initial implementation of hierarchical flags. This allows flags to be defined at multiple levels of a hierarchy, walking up through the list until finding an explicit setting. This is useful when the resolution of the value of a flag needs to be able to be overridden at multiple levels, such as a first checking a user, then a department, then an organization, and only then falling back tot he default.

The change is backwards compatible, using the existing actor protocols and requiring an explicit, new option of `for_hierarchy:` in place of `for:`.

The one bit of behaviour I haven't implemented yet is handling of actor percentages. I've gone back and forth on whether they should calculate the percentage on just the first actor in the list or on the collected set of all actors supplied. In other words, should

```elixir
FunWithFlags.enabled?(:feature, for_hierarchy: [user, division_a])
```
always give the same result as
```elixir
FunWithFlags.enabled?(:feature, for_hierarchy: [user, division_b])
```

In one sense it would make sense since you might always want a given actor to see the same behaviour across your system, regardless of the situation they're using it in. On the other...maybe not? Since it doesn't really come up in our particular use case, I'd welcome any input.


bernardd:
Whoops let me fix those credo warnings.

